PR #3, opened 2016-11-24
Make sys package build more robust

thijsc:
Remove `lib` from name to follow conventions and add sys packages for libz and lz4 to make sure they link properly.

I made a pull on the `lz4` crate to split out their build into a sys package. I will update this pull if that's merged: https://github.com/bozaro/lz4-rs/pull/16

This also removes the `Cargo.lock` files from the repo since they should not be commited for a library.


thijsc:
It looks like the Travis build is red because the git submodules haven't been loaded. Should we add this to `.travis.yml`?

thijsc:
I've uncommented the submodule initialization, that fixed the Travis build.

thijsc:
Good point. I did not realize that the full C code would be included when published as a crate.

I think checking `LICENCE` is a fine solution. That will allow the submodule to be initialized in CI and will also be reliable when used as a crate. I've force-pushed that change.

PR #4, opened 2016-11-25
Work with partitions

thijsc:


thijsc:
It'd also be a good option to just have a single `send_copy` method that has the partition option as the last argument.

fede1024:
Yes i would rather have one single method with an optional argument that can be None if no partition is specified (as you already did). I suggest to remove the second method, and to update the producer example.

thijsc:
Closing this one, will open separate ones based on the current `master`.

PR #6, opened 2016-11-28
Topic partition list and rebalance callback

thijsc:
The rebalance callback uses the same data structure that's used for managing subscriptions. This adds a wrapper around this structure which now is also used when subscribing.

Currently there's no way to get rebalance events in the streaming consumer. Not sure how to do that yet.


thijsc:
I will rebalance this once #7 has been merged.

thijsc:
Rebased again, now only contains it's own commits.

PR #7, opened 2016-11-28
Add lz4-sys crate

thijsc:
The separate `lz4-sys` crate was published, so we can re-add the dependency.

### DIFF
--- a/rdkafka-sys/Cargo.toml
+++ b/rdkafka-sys/Cargo.toml
@@ -10,9 +10,9 @@ description = "Native bindings to the librdkafka library"
 keywords = ["kafka", "rdkafka"]
 
 [dependencies]
-# lz4-sys = { git = "https://github.com/thijsc/lz4-rs.git" }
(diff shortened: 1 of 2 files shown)


rdkafka-sys/Cargo.toml
```diff
@@ -10,7 +10,7 @@ description = "Native bindings to the librdkafka library"
 keywords = ["kafka", "rdkafka"]
 
 [dependencies]
-# lz4-sys = { git = "https://github.com/thijsc/lz4-rs.git" }
+lz4-sys = ">= 1.0"
 libz-sys = ">= 1.0.7"
 openssl-sys = "> 0.7.0"
```
fede1024: Part of the inequality is missing (we don't want major version changes). We should probably use `^`, `~` or `*` instead.

thijsc:
I've made the dependencies stricter.

thijsc:
And also updated `openssl-sys` to the latest version.

PR #8, opened 2016-12-01
Tests

thijsc:
First stab at getting a test setup. It'd be nice to have Kafka running on Travis, but that's not currently the case yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: rust
+script:
+  - cargo build --verbose
+  - cargo test --lib --verbose
 
 notifications:
   webhooks:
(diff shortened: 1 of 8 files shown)


xitep:
just saw your comment, i think you might be interested in @dead10ck's experiments with kafka on travis; see [here](https://github.com/spicavigo/kafka-rust/issues/138) for more information; hope this helps.

thijsc:
Interesting. We should definitively use that approach once it works there. I'm following that issue now.

fede1024:
[Here](https://github.com/Yelp/kafka-utils) is a project i worked on that uses travis and docker compose to run automatic tests on Kafka 0.8 and 0.9. We could probably implement something similar for this project and for kafka-rust.

I was also wondering if we should implement the kafkatest test suite, like the go wrapper does ([link](https://github.com/confluentinc/confluent-kafka-go/tree/master/kafkatest)).

thijsc:
I've rebased on master

PR #10, opened 2016-12-04
Integrate rdkafka logging with log crate

thijsc:
Pipe rdkafka logs to Rust logging, so they show up in the same format and with the same levels as your normal logs.


thijsc:
I didn't find a trace level in `librdkafka` itself.

PR #13, opened 2016-12-09
Implement Error, Debug and Display for KafkaError

thijsc:
Implement these traits so it's possible to integrate `KafkaError` with normal Rust error handling and libraries such as `error-chain`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,6 +5,8 @@ extern crate std;
 
 use self::rdkafka::types::*;
 
+use util::cstr_to_owned;
+
 /// Kafka result
 pub type KafkaResult<T> = Result<T, KafkaError>;
 
@@ -26,7 +28,6 @@ impl IsError for RDKafkaConfRes {
     }
 }
 
-#[derive(Debug)]
 /// Represents all Kafka errors.
 pub enum KafkaError {
     ClientConfig((RDKafkaConfRes, String, String, String)),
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -41,8 +40,145 @@ pub enum KafkaError {
...
+    }
+}
+
 impl From<std::ffi::NulError> for KafkaError {
     fn from(err: std::ffi::NulError) -> KafkaError {
         KafkaError::Nul(err)
     }
 }
+
+fn rd_kafka_resp_err_description(err: RDKafkaRespErr) -> &'static str {
```
fede1024: Could we use the `rd_kafka_get_err_descs` function provided by librdkafka?

thijsc: Oh, didn't know that existed. Will take a look.

thijsc: You probably mean `rd_kafka_err2str`?

thijsc:
Changed the source of the error description to `rd_kafka_err2str`.

src/error.rs
```diff
@@ -41,8 +42,72 @@ pub enum KafkaError {
...
+    }
+}
+
 impl From<std::ffi::NulError> for KafkaError {
     fn from(err: std::ffi::NulError) -> KafkaError {
         KafkaError::Nul(err)
     }
 }
+
+fn rd_kafka_resp_err_description(err: RDKafkaRespErr) -> String {
```
fede1024: Since this is not a native rdkafka function, i would suggest renaming it. Something along the lines of `resp_err_description`.

thijsc:
Renamed the method and rebased. Did it on my master by accident, so it includes #14. Will rebase again once that's merged.

thijsc:
Rebased

PR #14, opened 2016-12-13
Remove log level from log callback

thijsc:
It's usually already displayed by the log implementation on the Rust side, so this results in duplicate output in the log.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -13,22 +13,21 @@ use client::Context;
 const ERR_LEN: usize = 256;
 
 unsafe extern "C" fn log_cb(
-    _: *const rdkafka::rd_kafka_s,
+    _client: *const RDKafkaState,
     level: i32,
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -15,20 +15,19 @@ const ERR_LEN: usize = 256;
 unsafe extern "C" fn log_cb(
     _: *const rdkafka::rd_kafka_s,
     level: i32,
-    fac: *const i8,
+    _: *const i8,
```
fede1024: I imagine the compiler will print a warning if you don't use the variable, right? If that's the case, can we rename it to `_fac` instead of just `_`? Same for the first one, that could be called `_client`.

Also: we could use the types defined in the rdkafka-sys::types module instead of rdkafka::rd_kafka_s.

thijsc:
Processed your feedback.

PR #16, opened 2016-12-19
Add pause and resume to consumer

thijsc:
Add pause and resume so you can pause consuming messages without triggering a rebalance.

### DIFF
--- a/rdkafka-sys/src/bindings.rs
+++ b/rdkafka-sys/src/bindings.rs
@@ -976,7 +976,7 @@ pub enum rd_kafka_queue_s { }
 pub type rd_kafka_queue_t = rd_kafka_queue_s;
 #[derive(Copy, Clone)]
 #[repr(i32)]
-#[derive(Debug)]
+#[derive(Debug,PartialEq)]
 pub enum rd_kafka_resp_err_t {
     RD_KAFKA_RESP_ERR__BEGIN = -200,
     RD_KAFKA_RESP_ERR__BAD_MSG = -199,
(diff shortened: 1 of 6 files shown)


rdkafka-sys/src/bindings.rs
```diff
@@ -976,7 +976,7 @@ pub enum rd_kafka_queue_s { }
 pub type rd_kafka_queue_t = rd_kafka_queue_s;
 #[derive(Copy, Clone)]
 #[repr(i32)]
-#[derive(Debug)]
+#[derive(Debug,PartialEq)]
```
thijsc: I added this here manually. Maybe there's an option in Bindgen to generate this? Not sure which command you used.

thijsc:
I've pushed a second commit that lets the streaming consumer handle the necessary polling automatically. This removes support for specifying the topics because this simplifies the process a lot. We could add this back in later if somebody needs it.

rdkafka-sys/src/bindings.rs
fede1024: I don't think there is any option to add it automatically via bindgen (command i used: `bindgen --builtins --convert-macros librdkafka-sys/librdkafka/src/rdkafka.h > outout.rs`).

We should be able to implement it directly, and we could add the implementation in src/types.rs. I've tried with the following snippet of code and the compiler doesn't complain:

```rust
impl PartialEq for RDKafkaRespErr {                                                                                                                                                                                                             
    fn eq(&self, other: &RDKafkaRespErr) -> bool {
        self == other
    }
}
```

thijsc:
Closing this one for now, I think the approach should be slightly different on second thought.

PR #17, opened 2016-12-23
Make fields in DeliveryStatus public

thijsc:
I need them to gather stats on wether producing is working.


fede1024:
Could you explain why this is needed, given that we have getters?

PR #21, opened 2017-01-02
Return a result for a DeliveryReport

thijsc:
This makes it possible to match on the report instead of resorting to using `is_error` and `resp_err_description` manually.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -17,7 +17,7 @@ use self::futures::{Canceled, Complete, Future, Poll, Oneshot};
 
 use client::{Client, Context};
 use config::{ClientConfig, FromClientConfig, FromClientConfigAndContext, TopicConfig};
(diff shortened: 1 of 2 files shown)


fede1024:
Yes this is definitely better than using than forcing the user to import and use the `is_error` function. Thanks!

src/producer.rs
```diff
@@ -153,9 +153,13 @@ impl DeliveryReport {
         }
     }
 
-    /// Returns the error associated with the production of the message.
-    pub fn error(&self) -> RDKafkaRespErr {
-        self.error
+    /// Returns the result of the production of the message.
+    pub fn result(&self) -> Result<(), KafkaError> {
```
fede1024: You can use `KafkaResult<()>` here.

thijsc:
Changed the return to `KafkaResult`

PR #22, opened 2017-01-02
Convert unwraps to expects and test for custom product context

thijsc:
This converts some unwraps to expects so I could rule out some causes of a crash I'm seeing. It looks like the crash is caused by using a custom producer context, so I've added an integration test for this.

The test in this pull runs indefinitely. When you change line 54 back to this:

```rust
.create::<FutureProducer<_>>()
```

The test passes within a second or two again. Seems like something breaks when using `create_with_context` instead of `create`, so far I have no idea why. Any thoughts?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -75,7 +75,7 @@ impl ClientConfig {
             if let Err(e) = topic_config {
                 return Err(e);
             } else {
-                unsafe { rdkafka::rd_kafka_conf_set_default_topic_conf(conf, topic_config.unwrap()) };
+                unsafe { rdkafka::rd_kafka_conf_set_default_topic_conf(conf, topic_config.expect("No topic config present when creating native config")) };
             }
         }
         unsafe { rdkafka::rd_kafka_conf_set_log_cb(conf, Some(log_cb)) };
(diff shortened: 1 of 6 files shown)


fede1024:
Thanks! Because of a bug in the `delivery_cb` function, the context gets freed every time a message it receives. The `FutureProducerContext` is just a wrapper for a generic `ProducerContext`, normally empty, so there is no double free unless a custom context is specified.

I'm pushing the fix.

fede1024:
Fix: https://github.com/fede1024/rust-rdkafka/commit/3a414227ae619954e1a07cf84107d9225a63222c

fede1024:
The way the `FutureProducer` allows to specify a `ProducerContext` is very confusing at the moment, since the only method currently available in the trait is not actually executed (since it is used by the `FutureProducer` itself). I think we should remove this test, since it is a bit misleading.

I'll update the `FutureProducer` to use a different trait instead of `ProducerContext` (or nothing at all).

thijsc:
Closing this one. I've cherry-picked the expect commit to #24 and opened #23 to track the producer context issue.

PR #26, opened 2017-01-06
Fix tests & commit list of partitions and offsets

thijsc:
We needed more control over committing in our project, so I added the other commit function from `rdkafka` too.


thijsc:
Hang on, need to correct a mistake.

thijsc:
Done, it now works properly.

PR #27, opened 2017-01-10
Pause and resume consumers

thijsc:
Pause and resume for both the base and streaming consumer.

The streaming consumer starts a separate thread that polls while the consumer is paused. This ensures that the consumer will be considered alive by Kafka even when we're not accepting the future on the other end.

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -81,6 +81,26 @@ impl<C: ConsumerContext> BaseConsumer<C> {
         Ok(())
     }
 
+    /// Pause consuming of this consumer.
+    pub fn pause(&self) {
+        unsafe {
+            let mut tp_list = rdkafka::rd_kafka_topic_partition_list_new(0);
+            rdkafka::rd_kafka_assignment(self.client.native_ptr(), &mut tp_list);
+            rdkafka::rd_kafka_pause_partitions(self.client.native_ptr(), tp_list);
+            rdkafka::rd_kafka_topic_partition_list_destroy(tp_list);
+        }
+    }
+
+    /// Resume consuming of this consumer.
+    pub fn resume(&self) {
+        unsafe {
+            let mut tp_list = rdkafka::rd_kafka_topic_partition_list_new(0);
+            rdkafka::rd_kafka_assignment(self.client.native_ptr(), &mut tp_list);
+            rdkafka::rd_kafka_resume_partitions(self.client.native_ptr(), tp_list);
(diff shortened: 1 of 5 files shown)


thijsc:
Turns out it doesn't need a background thread. @edenhill said:

> no, that's a Java thing. Librdkafka keeps the group membership alive thru background threads

I'll update this pull.

thijsc:
He also said:

> You will need to call consumer_poll() at regular intervals anyway so that broker-triggered rebalances are propagated and handled in a timely manner

I think this means that in the case of a consumer the polling thread is still necessary.

thijsc:
Not really sure what the correct approach is here. Will get back on this later.

thijsc:
Re-opening. I was able to use it like this in our project, so the basic idea works.

thijsc:
@fede1024 Any feedback on this?

fede1024:
Sorry for the delay, I'll review asap.

src/consumer/base_consumer.rs
```diff
@@ -81,6 +81,26 @@ impl<C: ConsumerContext> BaseConsumer<C> {
...
+            rdkafka::rd_kafka_pause_partitions(self.client.native_ptr(), tp_list);
+            rdkafka::rd_kafka_topic_partition_list_destroy(tp_list);
+        }
+    }
+
+    /// Resume consuming of this consumer.
+    pub fn resume(&self) {
+        unsafe {
+            let mut tp_list = rdkafka::rd_kafka_topic_partition_list_new(0);
+            rdkafka::rd_kafka_assignment(self.client.native_ptr(), &mut tp_list);
```
thijsc: I'm looking into a possible bug here at the moment: It seems like the consumer doesn't continue from the correct offsets.

thijsc:
1. I discussed this with @edenhill and he mentioned that when you're paused you should still poll to make sure that rebalance callbacks get processed.

2. What I was trying to avoid is the original loop being stuck on waiting on the future. It then wouldn't poll in the meantime and everything would stall. I'm not 100% sure this can actually happen, so maybe the main loop would be fine.

thijsc:
Discussed some more  in the meantime. We don't need the background threads because heartbeats are done in a separate background thread. This also means we don't have a direct use case for it anymore. Closing since I don't want to propose adding stuff we haven't used in production.

fede1024:
Right, we should still poll the consumer for the callbacks, i forgot about that. Let me know if you still need this change in the future. We could reopen the PR, remove the additional loop and merge.

PR #28, opened 2017-01-21
More metadata

thijsc:
Add various metadata that is available for a consumer.

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -81,13 +81,6 @@ impl<C: ConsumerContext> BaseConsumer<C> {
         Ok(())
     }
 
-    /// Returns a list of topics or topic patterns the consumer is subscribed to.
-    pub fn get_subscriptions(&self) -> TopicPartitionList {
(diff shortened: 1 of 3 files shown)


src/consumer/base_consumer.rs
```diff
@@ -121,6 +114,76 @@ impl<C: ConsumerContext> BaseConsumer<C> {
...
 
+    /// Returns the current topic subscription.
+    pub fn subscription(&self) -> KafkaResult<TopicPartitionList> {
+        let mut tp_list = unsafe { rdkafka::rd_kafka_topic_partition_list_new(0) };
+        let error = unsafe {
+            rdkafka::rd_kafka_subscription(self.client.native_ptr(), &mut tp_list)
+        };
+
+        if error.is_error() {
+            Err(KafkaError::MetadataFetch(error))
```
fede1024: `tp_list` should be freed before returning from this branch (same in the other methods).

thijsc:
Added the frees on error.

PR #32, opened 2017-02-09
Timestamps

thijsc:
Current status:

* [x] Add timestamp accessor to message
* [x] Upgrade bindings (you might want to do this again on Linux)
* [x] Convert produce function to varargs version
* [x] Actually add option to set timestamp when producing
* [x] Add integration test with timestamps

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -61,7 +61,8 @@ impl NativeTopic {
         partition: Option<i32>,
         payload: Option<&[u8]>,
         key: Option<&[u8]>,
-        delivery_context: Option<Box<T>>
+        delivery_context: Option<Box<T>>,
+        timestamp: Option<i64>
     ) -> KafkaResult<()> {
         let (payload_ptr, payload_len) = match payload {
             None => (ptr::null_mut(), 0),
@@ -75,6 +76,10 @@ impl NativeTopic {
             Some(context) => Box::into_raw(context) as *mut c_void,
             None => ptr::null_mut(),
         };
+        let timestamp = match timestamp {
+            Some(t) => t,
+            None => 0
+        };
         let produce_error = unsafe {
             rdsys::rd_kafka_producev(
                 self.client_ptr,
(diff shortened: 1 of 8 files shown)


fede1024:
I think i'll pick this branch manually and do some changes on top, before merging, if that's fine for you. What i have in mind is:

- Creating bindings for both linux and mac in two different files, and configure the building process to use the correct one depending on the OS.
- Adding the option to add the timestamp.
- Add integration tests.

thijsc:
> Creating bindings for both linux and mac in two different files, and configure the building process to use the correct one depending on the OS.

I don't think this is necessary. They do work, they're just different. We could also consider using this approach: http://fitzgeraldnick.com/2016/12/14/using-libbindgen-in-build-rs.html

fede1024:
I saw that blogpost few weeks ago, but my understanding is that to generate bindings at compile time, clang and llvm are required. At the moment building rust-rdkafka only requires a normal gcc setup (and rust, of course). I'm a bit reluctant to add additional requirements, but I'll read some more and check if my understanding is correct.

In the same blogpost they mention having to maintain multiple FFI bindings for different platforms (at the moment it seem to work anyway, but it might not be the case for all API calls and for future versions).

thijsc:
I can confirm that the Linux-generated bindings work on my Mac.

thijsc:
I finished this and rebased it on master.

src/producer.rs
```diff
@@ -26,7 +26,8 @@ use util::cstr_to_owned;
 
 /// A native rdkafka-sys topic, used for message production.
 struct NativeTopic {
-    ptr: *mut RDKafkaTopic,
+    client_ptr: *mut RDKafka,
```
fede1024: Including the `client_ptr` here is dangerous, since we might access a pointer that has been freed by another section of the code (it can't happen with `topic_ptr` since it's freed in the `Drop` method).

The right approach is probably to have a real reference or `Arc` to the client. I'll merge the change and then see how I can change that bit. It will probably require a bit of refactoring.

thijsc: This is actually not really a change, I only renamed the field to make it's meaning more clear.

fede1024: This line now contains a pointer to an `RDKafka`, which wasn't there before. However it's not a big deal, i'm working on a branch that reorganizes the way this part of the code works. The new version should be more straightforward.

PR #36, opened 2017-02-28
Parse statistics to struct

thijsc:
We could consider making the field names more self-explanatory and adding some docs.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -7,12 +7,14 @@ use std::mem;
 use std::os::raw::c_void;
 use std::ptr;
 
+use serde_json;
+
 use config::{ClientConfig, NativeClientConfig, RDKafkaLogLevel};
 use error::{IsError, KafkaError, KafkaResult};
 use groups::GroupList;
 use metadata::Metadata;
+use statistics::Statistics;
 use util::bytes_cstr_to_owned;
-
 
 /// A Context is an object that can store user-defined data and on which callbacks can be
 /// defined. Refer to the list of methods to see which callbacks can currently be overridden.
@@ -32,10 +34,10 @@ pub trait Context: Send + Sync {
         }
     }
 
-    /// Receives the statistics of the librdkafka client in JSON format. To enable, the
+    /// Receives the statistics of the librdkafka client. To enable, the
     /// "statistics.interval.ms" configuration parameter must be specified.
-    fn stats(&self, json: String) {
-        println!("Client stats: {}", json);
+    fn stats(&self, statistics: Statistics) {
+        info!("Client stats: {:?}", statistics);
     }
 
     // NOTE: when adding a new method, remember to add it to the FutureProducerContext as well.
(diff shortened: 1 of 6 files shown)


src/client.rs
```diff
@@ -209,16 +211,19 @@ pub unsafe extern "C" fn native_stats_cb<C: Context>(
...
+
+    let statistics_json = String::from_raw_parts(json as *mut u8, json_len, json_len);
+    match serde_json::from_str(&statistics_json) {
+        Ok(stats) => (*context).stats(stats),
+        Err(e) => error!("Could not parse statistics json: {}", e)
+    }
+
     mem::forget(context);   // Do not free the context
 
     1 // librdkafka will not free the pointer
```
fede1024: I think we should return a 0 here, to make librdkafka free the pointer to the original string.

src/statistics.rs
```diff
@@ -0,0 +1,495 @@
...
+    pub consumer_lag: i64,
+    pub txmsgs: i64,
+    pub txbytes: i64,
+    pub msgs: i64,
+    pub rx_ver_drops: i64
+}
+
+#[derive(Deserialize,Debug)]
+pub struct Rebalance {
+    pub rebalance_age: i64,
```
fede1024: we already have a structure named Rebalance. can we call it differently?

fede1024: The same actually apply to many other data structures. Let's leave it as is, it's unlikely to be a problem in practice.

src/client.rs
thijsc: I'm not sure what the point of this return code is, I only changed it up to line 221.

thijsc:
@fede1024 do you think we should give the fields more human readable names?

src/client.rs
fede1024: Sorry, my mistake. Our rust code will take care of freeing the string in this case as well.

fede1024:
I think that using the same naming convention that librdkafka uses has the advantage that all the librdkafka documentation automatically applies to ours (plus, the names are a bit shorter).

What we should do is to copy the documentation from the librdkafka documentation to each field in each structure, so that the meaning of each value will appear in the rust docs.

Let me know if you prefer merging the change as-is and we'll take care of the documentation later on, or if you prefer to include it in this code review.

thijsc:
I'll add the docs too. The length of the names doesn't matter in Rust right? I'm still a bit on the fence. If we add the docs inline you wouldn't need the original docs and it would be nice to have readable field names.

thijsc:
Tested this on our staging environment and made some fields optional. It should now work properly.

src/statistics.rs
```diff
@@ -0,0 +1,495 @@
+use std::collections::HashMap;
+
+#[derive(Deserialize,Debug)]
+pub struct Statistics {
+    pub name: String,
+    #[serde(rename = "type")]
```
fede1024: Why this renaming?

thijsc: `type` is a reserved word in Rust.

fede1024: /facepalm/ yes of course :)

PR #40, opened 2017-03-28
Termination fixes

thijsc:
Make some changes in behaviour to make sure we implement this properly:

https://github.com/edenhill/librdkafka/wiki/Proper-termination-sequence

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -88,7 +88,6 @@ impl Drop for NativeClient {
         trace!("Destroy rd_kafka {:p}", self.ptr);
         unsafe {
             rdsys::rd_kafka_destroy(self.ptr);
-            rdsys::rd_kafka_wait_destroyed(1000);
         }
     }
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -88,7 +88,6 @@ impl Drop for NativeClient {
         trace!("Destroy rd_kafka {:p}", self.ptr);
         unsafe {
             rdsys::rd_kafka_destroy(self.ptr);
-            rdsys::rd_kafka_wait_destroyed(1000);
```
fede1024: Why was this removed?

thijsc: See the commit message:

> It's not needed anymore in 0.9, see:
>
> https://github.com/edenhill/librdkafka/wiki/Proper-termination-sequence

fede1024: I didn't notice that. Thanks!

PR #41, opened 2017-04-13
Add offsets_for_timestamp to consumer

thijsc:
This allows you to look up the first offset for a certain timestamp.

It is possible to have a separate timestamp per partition, but we don't need this for our use case. We could add a `offsets_for_timestamps` that takes a map of partition/timestamp as well later if the need arises.


thijsc:
Build is red because rustup is down at the moment, will retry it later.

PR #45, opened 2017-05-12
Make topic partition list optional for commit on base consumer

thijsc:
If the pointer passed into rdkafka is null it will commit the current state. This is reliable on the base consumer.


fede1024:
Please add a line in the changelog.

thijsc:
Updated the changelog.

PR #46, opened 2017-05-12
Manual control of offsets for next commit

thijsc:
Add `store_offset` to consumer so you can manually control the offsets for the next commit.


src/consumer/stream_consumer.rs
```diff
@@ -20,7 +20,8 @@ use std::time::Duration;
 /// be polled and it will return all consumed messages as a `Stream`.
 /// Due to the asynchronous nature of the stream, some messages might be consumed by the consumer
 /// without being processed on the other end of the stream. If auto commit is used, it might cause
-/// message loss after consumer restart. Manual offset commit should be used instead.
+/// message loss after consumer restart. Manual offset storing should be used, see the `store_offset`
+/// function on `Consumer`.
```
thijsc: If you use the manual offset store approach autocommit is fine for the stream consumer, it will use the offsets you set yourself.

PR #47, opened 2017-05-17
Error callback

thijsc:
Add `error` callback to the `Context` so you can add behaviour for handling global errors in the client.


src/client.rs
```diff
@@ -277,6 +282,23 @@ pub unsafe extern "C" fn native_stats_cb<C: Context>(
     1 // librdkafka will not free the json buffer
 }
 
+pub unsafe extern "C" fn native_error_cb<C: Context>(
+        _client: *mut RDKafka, err: i32, reason: *const i8,
+        opaque: *mut c_void) {
+    let err = rdsys::primive_to_rd_kafka_resp_err_t(err).expect("global error not an rd_kafka_resp_err_t");
+    let error = match err {
+        rdsys::rd_kafka_resp_err_t::RD_KAFKA_RESP_ERR__ALL_BROKERS_DOWN => {
+            KafkaError::AllBrokersDown(err)
```
fede1024: Adding a special case for "all brokers down" makes the error handling less linear. People can already catch it if they want without having to add a specific error. Let's just use Global(err).

thijsc: It's actually pretty hard to match on, you need to get into the `rd_kafka_resp_err_t` values in rdsys. We did a similar thing with partition eof.

Maybe we should keep a mapping to a nice Rust enum of all values including documentation?

thijsc: Any thoughts on this?

fede1024: Sorry for the delay. I'd like to avoid having special cases such as `AllBrokersDown` if possible, but I don't know what is the best way to export them or define them so that matching specific errors becomes easy. I'll merge the branch and run some experiments on it.

PR #49, opened 2017-06-11
Remove explicit lifetime for Message

thijsc:
It's not needed because Rust will infer the correct lifetime. Leaving it out makes it easier to work with messages from client code. Also adds `Clone` and `Copy` derives to `Timestamp`.

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -22,7 +22,7 @@ pub struct Message {
 
 unsafe impl Send for Message {}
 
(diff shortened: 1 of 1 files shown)


thijsc:
Not sure why the build is failing, the tests are green on my machine.

fede1024:
One of the tests checks the time at which some events are processed, so it might be flaky. Your change looks good to me.

PR #74, opened 2018-02-11
WIP: Use Option<Duration> for timeouts

vorner:
Using i32 for milliseconds, with special case of -1 as no timeout, is
error prone and less self-documenting.

I tried to replace all externally-looking timeouts with the proper timeout encapsulation. What I'm not entirely sure of:
* If all them allow the `None`/-1 case for infinite timeout (I guess they probably do, but I haven't gone reading the rdkafka docs)
* That I got them all ‒ it was more of a search&replace than reading it manually from start to end.
* I also thought about taking `T: Into<Option<Duration>>` as the parameter. That would allow calling `poll(Duration::from_secs(10)` as well as `poll(None)`. While this is a bit less typing, it is also potentially bit more confusing and it would make certain traits not object safe, so I left it with this.

### DIFF
--- a/examples/metadata.rs
+++ b/examples/metadata.rs
@@ -7,10 +7,12 @@ use clap::{App, Arg};
 use rdkafka::consumer::{BaseConsumer, Consumer};
 use rdkafka::config::ClientConfig;
 
+use std::time::Duration;
+
 mod example_utils;
 use example_utils::setup_logger;
 
-fn print_metadata(brokers: &str, topic: Option<&str>, timeout_ms: i32, fetch_offsets: bool) {
+fn print_metadata(brokers: &str, topic: Option<&str>, timeout: Duration, fetch_offsets: bool) {
     let consumer = ClientConfig::new()
         .set("bootstrap.servers", brokers)
         .create::<BaseConsumer<_>>()
@@ -18,7 +20,7 @@ fn print_metadata(brokers: &str, topic: Option<&str>, timeout_ms: i32, fetch_off
 
     trace!("Consumer created");
 
-    let metadata = consumer.fetch_metadata(topic, timeout_ms)
+    let metadata = consumer.fetch_metadata(topic, timeout)
         .expect("Failed to fetch metadata");
 
     println!("Cluster information:");
@@ -43,7 +45,7 @@ fn print_metadata(brokers: &str, topic: Option<&str>, timeout_ms: i32, fetch_off
                      partition.isr(),
                      partition.error());
             if fetch_offsets {
-                let (low, high) = consumer.fetch_watermarks(topic.name(), partition.id(), 1000)
+                let (low, high) = consumer.fetch_watermarks(topic.name(), partition.id(), Duration::from_secs(1))
                     .unwrap_or((-1, -1));
                 println!("       Low watermark: {}  High watermark: {}", low, high);
             }
@@ -74,17 +76,20 @@ fn main() {
              .takes_value(true))
         .arg(Arg::with_name("timeout")
              .long("timeout")
-             .help("Metadata fetch timeout in seconds")
+             .help("Metadata fetch timeout in milliseconds")
              .takes_value(true)
-             .default_value("60.0"))
+             .default_value("60000"))
         .get_matches();
 
     setup_logger(true, matches.value_of("log-conf"));
 
     let brokers = matches.value_of("brokers").unwrap();
-    let timeout = value_t!(matches, "timeout", f32).unwrap();
+    let timeout = value_t!(matches, "timeout", u64).unwrap();
     let topic = matches.value_of("topic");
     let fetch_offsets = matches.is_present("offsets");
 
-    print_metadata(brokers, topic, (timeout * 1000f32) as i32, fetch_offsets);
+    print_metadata(brokers,
+                   topic,
+                   Duration::from_millis(timeout),
+                   fetch_offsets);
 }
--- a/src/client.rs
+++ b/src/client.rs
@@ -7,6 +7,7 @@ use std::slice;
 use std::mem;
 use std::os::raw::c_void;
 use std::ptr;
+use std::time::Duration;
 
 use serde_json;
 
@@ -15,7 +16,7 @@ use error::{IsError, KafkaError, KafkaResult};
 use groups::GroupList;
 use metadata::Metadata;
 use statistics::Statistics;
-use util::bytes_cstr_to_owned;
+use util::{bytes_cstr_to_owned, timeout_to_ms};
 
 /// A Context is an object that can store user-defined data and on which callbacks can be
 /// defined. Refer to the list of methods to see which callbacks can currently be overridden.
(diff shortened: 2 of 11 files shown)


fede1024:
I checked the rdkafka docs and the behavior is not always documented explicitly. However everywhere I checked (metadata fetch, offset fetch etc) the timeout is converted using [this method](https://github.com/edenhill/librdkafka/blob/8c67e42c0f694b83f4d92f15a6d508c6a95a2f9d/src/rdtime.h#L126). The only exception is poll which doesn't use it, but explicitly says that -1 means infinite and 0 means non-blocking.

This seems to suggest that your point n.1 holds, even though methods that don't have cached data might always fail if used in a non-blocking way (`rd_kafka_list_groups` for example explicitly says that the value must be > 0; we should test what happens when 0 or -1 are used, but I suspect that it will always timeout and potentially wait forever respectively).

What if instead of using an `Optional<Duration>`, we create a type that includes a "no wait" and a "infinite" wait time? Something like `Timeout::Millis(...)` or `Timeout::NoWait`. We could also have `Duration`s implicitly or explicitly converted to this type (using `Into` or `Timeout::from_duration(...)`). Let me know what you think about this.

I'm not sure I completely understand your object safety concerns in n.3.

vorner:
> What if instead of using an Optional<Duration>, we create a type that includes a "no wait" and a "infinite" wait time? Something like Timeout::Millis(...) or Timeout::NoWait. We could also have Durations implicitly or explicitly converted to this type (using Into or Timeout::from_duration(...)). Let me know what you think about this.

I don't see what this would bring, over `Optional<Duration>`. Few points about it:

* While the infinite is a strange special case (so using -1 feels wrong), 0 isn't too different from 1 or from 2. Therefore, I don't think it needs special handling. The fact it needs special handling in the library is probably just a implementation detail.
* I kind of see the advantage of talking about time in Rust-native types designed for time. Why invent another (less potent and integrated) time type? And why remove the units from the number (eg. `Timeout::Millis` uses just a number, while it still hints at the units).
* I probably could see some reason for `Timeout::Finite(Duration)` and `Timeout::Infinite`, but `None` sounds like „there's no timeout“ (not an infinite one), which looks also correct. We probably *could* want to have `Timeout::Relative(Duration)`, `Timeout::Absolute(Instant)` and `Timeout::Infinite`, but it looks a bit of an overkill.

About object safety ‒ I modified some trait on the way. A trait can't be made into a trait object (eg. erase the real type and use just `Box<Trait>`) if the trait has some generic methods. Having a `Timeout: Into<Option<Duration>>` is generic.

fede1024:
I see your point in not having a special case for `0` since `Optional<Duration>` is expressive enough. I was hoping we could find a way to make the API less verbose, as in most cases I find myself having a hardcoded timeout, which now requires writing `Some(Duration::from_millis(200))` (still, a minor issue).

This is definitely a step in the good direction though. Maybe we'll find a shorter but still idiomatic way to express the same parameter before releasing a new version of the library.

examples/metadata.rs
```diff
@@ -86,5 +88,8 @@ fn main() {
     let topic = matches.value_of("topic");
     let fetch_offsets = matches.is_present("offsets");
 
-    print_metadata(brokers, topic, (timeout * 1000f32) as i32, fetch_offsets);
+    print_metadata(brokers,
+                   topic,
+                   Some(Duration::from_millis((timeout * 1000f32) as u64)),
```
fede1024: I think we could just change the command line the parameter to milliseconds (directly in `u64`) and remove the conversion here.

src/client.rs
```diff
@@ -153,14 +154,15 @@ impl<C: Context> Client<C> {
...
-    pub fn fetch_metadata(&self, topic: Option<&str>, timeout_ms: i32) -> KafkaResult<Metadata> {
+    pub fn fetch_metadata(&self, topic: Option<&str>, timeout: Option<Duration>) -> KafkaResult<Metadata> {
         let mut metadata_ptr: *const RDKafkaMetadata = ptr::null_mut();
         let (flag, native_topic) = if let Some(topic_name) = topic {
             (0, Some(self.native_topic(topic_name)?))
         } else {
             (1, None)
         };
         trace!("Starting metadata fetch");
+        let timeout_ms = timeout_to_ms(timeout);
```
fede1024: nit: if `timeout_ms` is only used once, could we use the `timeout_to_ms` function directly instead? (same in the other methods)

vorner:
I had little time, I'll do the other changes soon. Also, I'd like to squash the fixups before merge.

About the less verbose API: I'm wondering, how useful/expected it would be for people to use `Box<Consumer<EmptyConsumerContext>>` instead of a concrete type ‒ is it likely to have both the BaseConsumer and StreamConsumer in the same application and use them interchangeably? And if so, would exchanging slightly longer calls on the boxed trait-object for shorter timeout API in the common case make sense?

fede1024:
I think it's very unlikely that users might want to use different type of consumers interchangeably, since the API for consuming messages is very different. I am wondering if someone might want to do it for testing, but that seems unlikely as well.

vorner:
How about now? (I'll solve the conflicts once it looks good)

fede1024:
Looks good. Thanks! :) I'm still not completely sold on the `Timeout` type being an overkill, so long as we make sure not to reinvent the wheel when it comes to storing time. I think there are enough methods requiring it and handling it the same way that it might be worth considering it.

Something like:
``` rust
enum Timeout {
    Finite(Duration),
    Infinite
}
```

For non blocking operations, we could have
``` rust
consumer.fetch_watermarks(topic, partition, Duration::from_secs(0))
// vs
consumer.fetch_watermarks(topic, partition, Timeout::no_block())
// or
consumer.fetch_watermarks(topic, partition, Timeout::from_secs(0))
```
The main advantage in this case would be that the meaning of the parameter would be self explanatory (it's clearly a timeout). Also, it doesn't require any generic type in the API.

For infinite timeout:
``` rust
consumer.fetch_watermarks(topic, partition, None)
// vs
consumer.fetch_watermarks(topic, partition, Timeout::Infinite)
```
I think in this example the second line is much easier to read without having to read any documentation or going through the code.

For fixed timeouts:
``` rust
consumer.fetch_watermarks(topic, partition, Duration::from_secs(10))
// vs
consumer.fetch_watermarks(topic, partition, Timeout::from_secs(10))
// or
consumer.fetch_watermarks(topic, partition, Duration::from_secs(10).into())
```
This one might feel like reinventing the wheel, however we are just wrapping a `Duration`. I don't think it has any real downside, as long as we also implement transformations from `Duration` to make conversions from one to the other as simple as possible.

Implementing a `Timeout` type might be beyond the scope of this PR. I would be fine with merging this and then experiment with a `Timeout` type. Let me know if you think there are major downsides in using it. Going back to your 3 points:
1) It's true that `None` could be synonym of `Timeout::Infinite`, but the second one i believe is more self-documenting (`None` could also be interpreted as zero).
2) True, there is no reason to have `Timeout::Millis`. As long as we can easily convert from `Duration`, users should be able to do what they want with normal Rust types and only use `.into()` or `Timeout::from(duration)` at the very end.
3) Another advantage which might make `Timeout` not an overkill anymore is avoiding generic types.

Sorry for the back and forth on this :) designing good APIs is hard! Please let me know what you think.

vorner:
I kind of still don't feel like a new `Timeout` type is needed or appropriate:

* This way, every single crate could/would be making their own `Timeout` type. These would be incompatible with each other. I don't think this crate is responsible to defining timeout types, it should be doing communication with the broker ‒ it seems out of scope for the crate.
* The `Option<Duration>` is actually used as a timeout in the [standard library](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_read_timeout).
* Maybe it is a small distinction, but I feel like „infinite timeout“ is a different thing than „no timeout“ ‒ the first feels like waiting a really long (infinitely long), but the second sounds like the clock isn't ticking at all. The first feels like it could actually time out during the judgement day.
* Furthermore, your analysis misses one case ‒ computed timeout. If I watch other events as well, then it's

```rust
consumer.fetch_watermarks(topic, partition, next_event - Instant::now())
// vs
consumer.fetch_watermarks(topic, partition, (next_event - Instant::now()).into())
```

fede1024:
I see. OK sounds good :) Let me know when I can merge.

vorner:
OK, squashed and rebased onto master to get rid of the conflicts.

PR #75, opened 2018-03-02
Reexport DeliveryFuture from the producer module

fralalonde:
It seems that `DeliveryFuture` was left out from the reexported producer structs. This change fixes this omission, making possible to use `FutureProducer::send_copy_result()` by using just `rdkafka::producer::*` rather than having to include the struct from the `future_producer` module.


fede1024:
Thanks!

PR #76, opened 2018-03-06
Iterator interface for the consumer

vorner:
Allow iterating over the consumer instead of calling poll to get the
messages.

I didn't find any test for the base consumer ‒ should I try creating one, to test the iterator?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ## master
 
+* Added iterator interface to the `BaseConsumer`.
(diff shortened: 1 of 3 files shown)


vorner:
I added a fixup commit with the changelog and a test.

I hope the tests pass, though. I don't have a local kafka to test against yet :-(.

fede1024:
I run the test and it works :) (same for travis). I think it's good to merge, let me know if you prefer squashing first.

vorner:
Squashed, the diff is exactly the same.

Thank you :-)

PR #79, opened 2018-03-13
Simplify the iterator interface

vorner:
As the result of the talk we had online, this retries inside the iterator if the poll returns no message (it still returns the errors).

Also, got rid of the timeout, as it looks strange on the iterator and would terminate it „randomly“, which would probably be strange.

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -118,10 +118,9 @@ impl<C: ConsumerContext> BaseConsumer<C> {
 
     /// Returns an iterator over the available messages.
     ///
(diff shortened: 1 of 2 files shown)


src/consumer/base_consumer.rs
```diff
@@ -145,7 +144,7 @@ impl<C: ConsumerContext> BaseConsumer<C> {
     /// #    .create()
     /// #    .unwrap();
     ///
```
fede1024: A minor cosmetic improvement: I noticed that these empty lines will show up in [the docs](https://fede1024.github.io/rust-rdkafka/rdkafka/consumer/base_consumer/struct.BaseConsumer.html#method.iter). I suggest removing them.

Issue #89, opened 2018-07-03
Achieving committing offsets only after successful productions


blt:
I've got myself worked into a corner and I'm not sure how to work back out with the API as-is. I could use a little guidance. My basic ambition is laid out [in an issue on the python client](https://github.com/confluentinc/confluent-kafka-python/issues/300). That is, I'd like to write a process that reads from some `input` topic and writes an an `output` topic and commits offsets to `input` only if the record has been successfully produced to `output`. 

As of 0.17, rdkafka's [`ProducerContext::delivery`](https://docs.rs/rdkafka/0.17.0/rdkafka/producer/base_producer/trait.ProducerContext.html#tymethod.delivery) is called with a [`DeliveryResult`](https://docs.rs/rdkafka/0.17.0/rdkafka/message/type.DeliveryResult.html) that contains a [`BorrowedMessage`](https://docs.rs/rdkafka/0.17.0/rdkafka/message/struct.BorrowedMessage.html) to the record published to `output` and not the `BorrowedMessage` which came in from `input`. But, [`Consumer::store_offset`](https://docs.rs/rdkafka/0.17.0/rdkafka/consumer/trait.Consumer.html#method.store_offset) requires a `&BorrowedMessage` be passed, the same that came in from the `Consumer`. 

Okay, so I have to get the original `&BorrowedMessage` through to `ProducerContext::delivery` along with a reference to the `Consumer` to accomplish my aim, I think. It occurred to me that I might pass a `(&BorrowedMessage, &Consumer)` through as a `ProducerContext::DeliveryOpaque` but that seems excessively complicated. Not to mention, there will need to be a rectification layer in there somewhere so that offsets are only committed in-order. I'm _really_ not sure how to fit that in. 

I noticed that [`examples/at_least_once.rs`](https://github.com/fede1024/rust-rdkafka/blob/master/examples/at_least_once.rs) accomplishes this by auto-committing explicitly stored offsets but waits synchronously for produced records to signal back with success. This is close to what I'd like to accomplish but the throughput is a little disappointing, what with the serialization on every production. 

I guess, is there a more throughput oriented way to commit consumed offsets only after records that result from them have been successfully produced other than serializing on writes? It feels like there's something close with `ProducerContext::delivery` but I'm not quite finding it.

blt:
I spent a few days with rdkafka-sys and I wonder if an API adjustment for this crate wouldn't do the trick. `BaseConsumer::store_offset` is:

```
[...]

PR #90, opened 2018-07-09
Modify topic_mirror to include questions, notes

blt:
This commit adjusts the topic_mirror example to include retry
in the case of enqueuing failure, removes the TopicPartitionList
in favor of performing offset commits directly with BorrowedMessage.
It's possible I've not understood the purpose of TopicPartitionList.

As it is now, I think I've saved on one allocation per incoming message
by removing TopicPartitionList, have imposed additional allocations per
failed enqueuing and have open questions around retrying when production
fails.

Apologies for some of the rustfmt reshuffling. 

Relevant issue: [#89](https://github.com/fede1024/rust-rdkafka/issues/89)

Signed-off-by: Brian L. Troutwine <briant@unity3d.com>

### DIFF
--- a/examples/topic_mirror.rs
+++ b/examples/topic_mirror.rs
@@ -12,13 +12,14 @@ use rdkafka::Message;
 
 use std::collections::binary_heap::BinaryHeap;
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 mod example_utils;
 use example_utils::setup_logger;
 
 use rdkafka::consumer::ConsumerContext;
+use rdkafka::consumer::Rebalance;
 use rdkafka::error::KafkaError;
 use rdkafka::error::KafkaResult;
 use rdkafka::message::BorrowedMessage;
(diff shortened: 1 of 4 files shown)


examples/topic_mirror.rs
```diff
@@ -53,34 +52,61 @@ fn create_consumer(brokers: &str, group_id: &str, topic: &str) -> LoggingConsume
...
+                // Question(blt) -- This call to `store_offset` will eventually
+                // cause offsets to be committed out of order. That will cause
+                // data loss. Should, then, the author cobble together a MinHeap
+                // (or similar) approach to commit only when all the suitable
+                // offsets have come through -- implying storage here -- or rely
+                // on rdkafka to take care of this?
+                //
+                // Either approach, potentially, interacts with the retry
+                // question below.
+                while let Err(_) = self.source_consumer.store_offset(&m) {
```
fede1024: Here `m` is the message that was just sent to Kafka (the copy, not the original one), so you'd be committing the offset of the copy, which is not what you want. The `TopicPartitionList` stores the location (offset and partition) of the original message, which is the one you want to commit.

examples/topic_mirror.rs
```diff
@@ -53,34 +52,61 @@ fn create_consumer(brokers: &str, group_id: &str, topic: &str) -> LoggingConsume
...
-                self.source_consumer.store_offset_list(&delivery_opaque);
-            },
-            Err((e, m)) => error!("Failed to deliver to {}: {:?}", m.topic(), e)
+                debug!("Delivered copy of {:?} to {}", m.payload().unwrap(), m.topic());
+                // Question(blt) -- This call to `store_offset` will eventually
+                // cause offsets to be committed out of order. That will cause
+                // data loss. Should, then, the author cobble together a MinHeap
+                // (or similar) approach to commit only when all the suitable
+                // offsets have come through -- implying storage here -- or rely
+                // on rdkafka to take care of this?
```
fede1024: `librdkafka` doesn't provide a way to commit in order, so yes, we should implement a different way to keep track of them and only periodically commit the lowest one per partition.

examples/topic_mirror.rs
blt: Ah, d'oh. Of course.

examples/topic_mirror.rs
blt: Here, by rdkafka, I meant this crate, not `librdkafka`.

blt: Whether rdkafka would be the right place for this functionality or not, I'm still a touch confused about how to pull this off, mostly down to lifetime management. 

In this example, the `TopicPartitionList` will always have one originating message associated with it. My initial idea was to `TopicPartitionList::elements` to retrieve the single `TopicPartitionListElem` and store that in an `BinaryHeap`. There's no `Ord` impl for `TopicPartitionListElem` yet—resolvable—but the key difficulty is the lack of Sync on `*mut rdkafka_sys::bindings::macos_64::rd_kafka_topic_partition_list_s`. My latest commit has the full code changes to reproduce, but here's a sample of the compilation error:

```
error[E0277]: `*mut rdkafka_sys::bindings::macos_64::rd_kafka_topic_partition_list_s` cannot be shared between threads safely
  --> examples/topic_mirror.rs:70:10
   |
70 | impl<'a> ProducerContext for MirrorProducerContext<'a> {
   |          ^^^^^^^^^^^^^^^ `*mut rdkafka_sys::bindings::macos_64::rd_kafka_topic_partition_list_s` cannot be shared between threads safely
   |
   = help: within `rdkafka::topic_partition_list::TopicPartitionListElem<'a>`, the trait `std::marker::Sync` is not implemented for `*mut rdkafka_sys::bindings::macos_64::rd_kafka_topic_partition_list_s`
   = note: required because it appears within the type `rdkafka::TopicPartitionList`
   = note: required because it appears within the type `&'a rdkafka::TopicPartitionList`
   = note: required because it appears within the type `rdkafka::topic_partition_list::TopicPartitionListElem<'a>`
   = note: required because of the requirements on the impl of `std::marker::Sync` for `std::ptr::Unique<rdkafka::topic_partition_list::TopicPartitionListElem<'a>>`
   = note: required because it appears within the type `alloc::raw_vec::RawVec<rdkafka::topic_partition_list::TopicPartitionListElem<'a>>`
   = note: required because it appears within the type `std::vec::Vec<rdkafka::topic_partition_list::TopicPartitionListElem<'a>>`
   = note: required because it appears within the type `std::collections::BinaryHeap<rdkafka::topic_partition_list::TopicPartitionListElem<'a>>`
   = note: required because it appears within the type `MirrorProducerContext<'a>`
```

blt:
@fede1024 I've added in-order offset commits to my most recent commit. It's a little gross as-is and I'm not sure if the mutability modifications I've made to BaseProducer are acceptable. After doing this I have a vague notion that this might live best in a Consumer somewhere.

blt:
I've realized now that I've got this wired up backward: the consumer needs to be the one to keep the offset run minheap, not the producer. 

Consider the use case where the user is not trying to establish a mirror but do filtering in some fashion. There, the calling code will need to signal to the Consumer that an offset has been processed--via `store_offset` I guess--without involving the Producer. 

I'll adjust my PR appropriately.

blt:
Er, no. Just moving the TopicOffsetMap into Consumer suffers the same problem of needing to adjust mutability of the self. So, I started down the road of wrapping `LoggingConsumer` in a struct which would hold the TopicOffsetMap which could be mutable and expose the appropriate higher-level API for cutting the Producer out when needed. This approach is going to run into issues with rebalancing. It's possible to extend the ConsumerContext impl to update the TopicOffsetMap but that fights against the wrapping-struct approach.

If the ConsumerContext functions were adjusted to be `&mut self` then the wrapping struct could be abandoned for coping with rebalance events. So far so good. That leaves committing offsets in-order. At present there's a `ConsumerContext::commit_callback` which does post-commit work. What about adjusting that to be `post_commit_callback` and introducing a `pre_commit_callback`? The pre version would be responsible for properly setting the offsets in TopicPartitionList prior to commit, in this use case.

PR #91, opened 2018-07-12
Added method for storing multiple offsets

ilya-epifanov:


fede1024:
Thank you for the PR. In the `test/topic_mirror` branch we are introducing a similar method ([link](https://github.com/fede1024/rust-rdkafka/blob/test/topic_mirror/src/consumer/mod.rs#L171)), which works directly on a `TopicPartitionList`. The `TopicPartitionList` type is used in other parts of the library where we need to store this kind of information, so to keep the library consistent I believe it is better to keep using the same type everywhere. I can move that method to master if it would help your use case.

ilya-epifanov:
I've just moved those methods from your branch and dumped my first attempt:)

Thanks!

petercable:
It would be great to have this method on master so that we can use the rdkafka offset store functionality without having to have the original `BorrowedMessage`.

ilya-epifanov:
@fede1024 ?

src/consumer/base_consumer.rs
```diff
@@ -235,6 +235,15 @@ impl<C: ConsumerContext> Consumer<C> for BaseConsumer<C> {
         }
     }
 
+    fn store_offset_list(&self, tpl: &TopicPartitionList) -> KafkaResult<()> {
```
flavray: In an effort to stay closer to `librdkafka`'s naming schema, would it make sense to name the functions `store_offsets`?

PR #104, opened 2018-12-18
fix doc typo

niedhui:
simply fix doc typo


flavray:
Good catch. Thank you for contributing! 🙂

PR #106, opened 2019-02-04
Added `offsets_for_times` method

ilya-epifanov:
This method allows to specify a TPL for which to get offsets.
The main reason is to allow getting offsets for timestamps without/before subscribing to anything.

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -313,6 +313,25 @@ impl<C: ConsumerContext> Consumer<C> for BaseConsumer<C> {
         }
     }
 
+    fn offsets_for_times<T: Into<Option<Duration>>>(&self, mut timestamps: TopicPartitionList, timeout: T)
+                                                    -> KafkaResult<TopicPartitionList>
+    {
+        // This call will then put the offset in the offset field of this topic partition list.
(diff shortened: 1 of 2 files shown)


src/consumer/base_consumer.rs
```diff
@@ -313,6 +313,25 @@ impl<C: ConsumerContext> Consumer<C> for BaseConsumer<C> {
         }
     }
 
+    fn offsets_for_times<T: Into<Option<Duration>>>(&self, mut timestamps: TopicPartitionList, timeout: T)
```
flavray: Would it make sense to add a comment here indicating that the `timestamps` is expected to contain have their offset set to the timestamp to search for?

src/consumer/base_consumer.rs
```diff
@@ -313,6 +313,25 @@ impl<C: ConsumerContext> Consumer<C> for BaseConsumer<C> {
...
+                                                    -> KafkaResult<TopicPartitionList>
+    {
+        // This call will then put the offset in the offset field of this topic partition list.
+        let offsets_for_times_error = unsafe {
```
flavray: This code seems to be the same code used in `offsets_for_timestamp`. Would it make sense to refactor `offsets_for_timestamp` to call this new function?

PR #107, opened 2019-02-04
Added `committed_offsets` method

ilya-epifanov:
This method allows to specify a TPL for which to get committed offsets.

The main reason is to allow getting committed offsets without/before subscribing to anything.

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -284,6 +284,18 @@ impl<C: ConsumerContext> Consumer<C> for BaseConsumer<C> {
         }
     }
 
+    fn committed_offsets<T: Into<Option<Duration>>>(&self, mut tpl: TopicPartitionList, timeout: T) -> KafkaResult<TopicPartitionList> {
+        let committed_error = unsafe {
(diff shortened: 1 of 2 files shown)


src/consumer/base_consumer.rs
```diff
@@ -284,6 +284,18 @@ impl<C: ConsumerContext> Consumer<C> for BaseConsumer<C> {
         }
     }
 
+    fn committed_offsets<T: Into<Option<Duration>>>(&self, mut tpl: TopicPartitionList, timeout: T) -> KafkaResult<TopicPartitionList> {
+        let committed_error = unsafe {
```
flavray: This code seems to be the same as in `committed` (apart from the `.ptr()` which shouldn't be a problem). Would it make sense to refactor `committed` to call `committed_offsets`?

svanharmelen:
I know this is a bit old 🙈 But how can I get the offsets for all partitions from a topic using this method? If I try to call it with `tpl.add_topic_unassigned` I get `-1001` which seems to indicate there is no offset returned. But how (as a consumer) can I know which partitions to add?

PR #108, opened 2019-02-04
Made `tpl` in `rebalance` mutable

ilya-epifanov:
`tpl` should be mutable in the rebalance hook.

An example: https://github.com/edenhill/librdkafka/blob/bd74dc31ba6edf431749b4ad5a00aa4fc78882fc/tests/0029-assign_offset.c#L70

### DIFF
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -42,7 +42,7 @@ pub trait ConsumerContext: ClientContext {
         &self,
         native_client: &NativeClient,
         err: RDKafkaRespErr,
-        tpl: &TopicPartitionList,
+        tpl: &mut TopicPartitionList,
     ) {
 
         let rebalance = match err {
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -48,9 +48,9 @@ unsafe extern "C" fn native_rebalance_cb<C: ConsumerContext>(
     // let context: &C = &*(opaque_ptr as *const C);
     let context = Box::from_raw(opaque_ptr as *mut C);
     let native_client = NativeClient::from_ptr(rk);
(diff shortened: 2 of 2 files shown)


src/consumer/mod.rs
```diff
@@ -42,7 +42,7 @@ pub trait ConsumerContext: ClientContext {
         &self,
         native_client: &NativeClient,
         err: RDKafkaRespErr,
-        tpl: &TopicPartitionList,
+        tpl: &mut TopicPartitionList,
```
flavray: It looks to me like the current code is correct.

[`rd_kafka_assign`](https://github.com/edenhill/librdkafka/blob/bd74dc31ba6edf431749b4ad5a00aa4fc78882fc/src/rdkafka_subscription.c#L97) is expecting a `const` pointer. The code you pointed out uses code that mutates the value pointed to the pointer, however the pointer itself seems constant.

Is the current code causing issues when using it?

flavray: It looks to me like the current code is correct.

[`rd_kafka_assign`](https://github.com/edenhill/librdkafka/blob/bd74dc31ba6edf431749b4ad5a00aa4fc78882fc/src/rdkafka_subscription.c#L97) is expecting a `const` pointer. The code you pointed out uses code that mutates the value pointed to the pointer, however the pointer itself seems constant.

Is the current code causing issues when using it?

ilya-epifanov: It's not possible to mutate &TPL now, but it's not forbidden in librdkafka.
The definition of the callback: https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L1599

flavray: Fair enough, thank you for the patch 🙂

PR #110, opened 2019-03-04
examples/asynchronous_processing: use updated Tokio API

benesch:
The tokio::executor::{current_thread, thread_pool} exports have been
deprecated since Tokio v1.8. Use the newer runtime APIs instead. There
is no functionality change; this just makes the asynchronous_processing
exampleclearer for folks who are new to the Tokio ecosystem.

### DIFF
--- a/examples/asynchronous_processing.rs
+++ b/examples/asynchronous_processing.rs
@@ -7,8 +7,7 @@ extern crate tokio;
 
 use clap::{App, Arg};
 use futures::{Future, lazy, Stream};
-use tokio::executor::current_thread;
-use tokio::executor::thread_pool;
+use tokio::runtime::current_thread;
 
 use rdkafka::Message;
 use rdkafka::consumer::Consumer;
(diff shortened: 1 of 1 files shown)


examples/asynchronous_processing.rs
```diff
@@ -66,14 +65,14 @@ fn run_async_processor(brokers: &str, group_id: &str, input_topic: &str, output_
...
         .expect("Producer creation error");
 
-    // Create the thread pool where the expensive computation will be performed.
-    let thread_pool = thread_pool::Builder::new()
+    // Create the runtime where the expensive computation will be performed.
+    let mut thread_pool = tokio::runtime::Builder::new()
         .name_prefix("pool-")
-        .pool_size(4)
-        .build();
+        .build()
```
flavray: For the sake of parity with the previous example, would it make sense to keep the `.pool_size(4)` here?

benesch: Oh, yeah, looks like the option still exists, but it's been renamed to `core_threads`, which is why I didn't spot it. Updated!

PR #117, opened 2019-04-18
Upgrade librdkafka and auto-generate bindings

benesch:
This PR is the result of a deep rabbit hole to get rust-rdkafka to have an updated set of Kafka error codes. It's not quite ready to go, as it depends on a revision of librdkafka that's not yet been merged (https://github.com/edenhill/librdkafka/pull/2287), but I wanted to open to get the discussion/CI run going.

The biggest change here is the auto-generation of bindings, as generating updated FreeBSD bindings for the librdkafka bump seemed rather annoying. (I don't have easy access to a FreeBSD machine, and it sounds like you don't either, @fede1024.)

PR #119, opened 2019-04-23
Automatically generate bindings + upgrade to v1.0.0

benesch:
This is a slimmed-down version of #117 that doesn't included unmerged librdkafka patches. Should be ready for review! 🎉


benesch:
Ok, I think I tracked down the source of the Linux failures. The problem was the `--enable-static` flag, which actively hinders static compilation. The tl;dr is that `--enable-static` is meant for when you want a shared librdkafka.so that statically links libz/liblz, but that's a very unusual thing to do, and requires a libz.a that's been compiled with -fPIC, which is *not* true of the libz.a that ships with Ubuntu/CentOS/probably most Linux distributions. If you just want a plan ol' static librdkafka.a, like we do, the correct thing to do is to omit `--enable-static`.

flavray:
Great investigation on the linux build failure :)

PR #140, opened 2019-08-20
Remove rustfmt.toml, reformat with rustfmt

vorot93:
Most current rustfmt options have been removed anyway so it's better to just adopt the official [Rust style guide](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/guide.md).


benesch:
Thanks! Rebased and merged.

PR #142, opened 2019-08-24
Upgrade log libraries and minor syntax change

lynxluna:
Change log crates to these

- `log` to `0.4.8`
- `env_logger` to `0.6.2`

Change the `example_utils.rs` to be able to be compiled with those newer log libraries.

Also some syntax changes (removing semicolon and using `dyn` for trait object)


benesch:
Thanks, this looks great! I've pushed an update to this PR that bumps to the now-latest version of env-logger (0.7.1). Will merge on green.

PR #150, opened 2019-09-25
Bump bindgen to 0.51.1 in rdkafka-sys

### LINKED ISSUE #148: rdkafka-sys error: the trait `Copy` may not be implemented for this type
I encountered this error when building the master branch. It seems a [bindgen issue](https://github.com/rust-lang/rust-bindgen/issues/458) is relevant.

```
error[E0204]: the trait `Copy` may not be implemented for this type
    --> /path/to/crate/target/debug/build/rdkafka-sys-7db474a7fb7f90d0/out/bindings.rs:3460:17
     |
3460 | #[derive(Debug, Copy, Clone)]
     |                 ^^^^
...
3465 |     pub __cmsg_data: __IncompleteArrayField<::std::os::raw::c_uchar>,
     |     ---------------------------------------------------------------- this field does not implement `Copy`

```

sd2k:
Fixes #148.

The issue in #148 only seems to happen on recent nightly compilers. This PR compiles both on stable (1.37.0) and nightly.

### DIFF
--- a/rdkafka-sys/Cargo.toml
+++ b/rdkafka-sys/Cargo.toml
@@ -16,7 +16,7 @@ zstd-sys = { version = "1.3", features = [] }
 openssl-sys = { version = "~ 0.9.0", optional = true }
 
 [build-dependencies]
-bindgen = "0.49.2"
+bindgen = "0.51.1"
 num_cpus = "0.2.0"
 pkg-config = "0.3.9"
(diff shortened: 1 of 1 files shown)


benesch:
Thanks for the PR, @sd2k! I landed the bindgen bump as part of #151, so we should be good to go now.

PR #151, opened 2019-10-01
Bump librdkafka to 1.2.0

thijsc:


thijsc:
This is broken because of llvm issues on some Travis images, think I need to look into getting rid of that dependency :-(.

benesch:
Nah, that LLVM warning's a red herring. I think this is fixed by just upgrading the bindgen version (#150). I'm going to submit a PR that combines this and #151.

thijsc:
Closing in favor of #154

PR #155, opened 2019-10-04
Fix arm build again

jerry73204:
It fixes incompatibility with ARM target, tested on my aarch64 machine.
The incompatibility is due to incorrect usage of `c_char` types.
The PR is based on top of s2tk's [bump bindgen PR](https://github.com/fede1024/rust-rdkafka/pull/150). Please accept that PR before accepting this.


benesch:
Thanks for the PR, @jerry73204! I've incorporated your patch into #762 alongside a patch that ensures CI will prevent the ARM build from breaking in the future.

Issue #98, opened 2018-09-19
Transactions example for read-process-write use case. 


thedodd:
I’m basically looking for an example of how to achieve the same thing here: https://www.confluent.io/blog/transactions-apache-kafka/

Thanks for all the hard work on this crate.

thedodd:
After searching around a bit more, I see that the EOS effort is still underway. All good.

PR #164, opened 2019-10-23
Plumb message_queue_nonempty callback

benesch:
This makes it possible to receive a notification via a Rust callback when there are new messages available. The bulk of the implementation is in the last two commits; the first commit is just preparatory work.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -10,7 +10,7 @@ use crate::rdsys::types::*;
 use crate::client::{Client, ClientContext, DefaultClientContext, NativeQueue};
 use crate::config::{ClientConfig, FromClientConfig, FromClientConfigAndContext};
 use crate::error::{IsError, KafkaError, KafkaResult};
-use crate::util::{cstr_to_owned, timeout_to_ms, AsCArray, ErrBuf, IntoOpaque, WrappedCPointer};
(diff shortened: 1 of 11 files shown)


benesch:
@quodlibetor or @thijsc, can I talk one of you into reviewing this?

thijsc:
Wouldn't it be possible to use a callback on the future? Or does that have downsides for you?

benesch:
Yeah, at @MaterializeInc we're not using the futures consumer. We have our own event loop that calls `consumer.poll(0)` at regular intervals. When we don't receive any messages, we want to be able to put the thread with the event loop to sleep, but get woken up if a new message arrives, which this PR enables. (Without this PR, the only option is to create another thread that forwards received Kafka messages over a channel, which creates unnecessary overhead.)

benesch:
Come to think of it, it should be possible to avoid the separate thread in the StreamConsumer entirely with the functionality in this PR. I'm going to poke at that a bit.

thijsc:
This is blocked on getting the build green in #98.

benesch:
>  I'm going to poke at that a bit.

This turns out to be fairly difficult, so I'm just going to merge as is, now that tests are passing. (Just needed to increase the timeout for CI, because things seem to be a bit slower due to the small Travis VMs.)

Issue #153, opened 2019-10-02
Cannot trigger a rebalance without calling poll


emilk:
I am trying to implement exactly-once delivery using several consumers (running in different instances). Each message should only end up in exactly ONE consumer. When I start a new consumer I want to trigger an automatic rebalance so that the new consumer gets assigned a few partitions. The existing consumers commits on rebalance, so that the new consumer does not read the same messages (avoiding duplicates).

The problem is that calling [`subscribe`](https://docs.rs/rdkafka/0.21.0/rdkafka/consumer/trait.Consumer.html#method.subscribe) does NOT trigger a rebalance. The only method I have found which reliably triggers a rebalance is a blocking (`None` timeout) call to [`poll`](https://docs.rs/rdkafka/0.21.0/rdkafka/consumer/base_consumer/struct.BaseConsumer.html#method.poll). The problem is that this will read `max.partition.fetch.bytes` worth of messages that the existing consumers have already processed, thus creating duplicates.

Is there any other way to trigger a rebalance?

benesch:
Interesting! Upstream suggests that you pause the consumer, call poll until the rebalance, and then unpause it: https://github.com/edenhill/librdkafka/issues/2394

Unfortunately, we don't currently expose pause, but we could! PRs welcome.

PR #167, opened 2019-10-25
Implement pause/resume for Consumer

emilk:
Closes https://github.com/fede1024/rust-rdkafka/issues/153


benesch:
This is beautiful, thank you! I amended the test slightly to perform two fewer iterations, just in the interest of bringing the runtime of the test down by a few seconds.

PR #169, opened 2019-11-05
Add `ssl_vendored` feature

a-rodin:
This PR adds new feature `ssl_vendored`. It allows to use `vendored` feature from `openssl-sys` crate in order to not depend on system-wide SSL installation.

Some context about environment variables exported by build scripts of dependencies can be found [here](https://kornel.ski/rust-sys-crate#headers).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ default = []
 # Enable ssl and sasl support
 ssl = ["rdkafka-sys/ssl"]
 sasl = ["rdkafka-sys/sasl", "ssl"]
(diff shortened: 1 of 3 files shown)


a-rodin:
@benesch CMake uses `OPENSSL_ROOT_DIR` environment variable to determine location of OpenSSL, which is set before both normal and CMake builds, so after c7ee668 CMake builds are supported too.

benesch:
Sorry for the delay here! I folded this in to #169, which standardizes our support for exposing dependency's features. Thanks very much, @a-rodin!

PR #176, opened 2019-11-18
Support CMake build on MSVC

a-rodin:
Builds on Windows with MSVC environment produce the library located at `build/src/Release` or `build/src/Debug`, instead of just `build/src` for other environments.

This PR explicitly handles this case. The choice between `Release` and `Debug` happens using the same logic as in `cmake` crate, see [these lines](https://docs.rs/cmake/0.1.26/src/cmake/lib.rs.html#329-333).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,13 @@ jobs:
     rust: stable
   - os: osx
     rust: stable
+  - os: windows
+    arch: amd64
+    rust: stable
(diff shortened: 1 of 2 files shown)


benesch:
For posterity, this works around what appears to be a long standing bug in the CMake crate: https://github.com/alexcrichton/cmake-rs/issues/18

PR #183, opened 2019-11-30
Improve build system + prepare for 0.22

benesch:
Probably easiest just to look at the commit messages and the updates to the changelog/README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,33 +114,38 @@ Add this to your `Cargo.toml`:
 
 ```toml
 [dependencies]
-rdkafka = "~0.21"
+rdkafka = { version = "0.22", features = ["cmake-build"] }
 ```
 
-This crate will compile librdkafka from sources and link it statically to your executable. To compile librdkafka you'll need:
+This crate will compile librdkafka from sources and link it statically to your
+executable. To compile librdkafka you'll need:
 
 * the GNU toolchain
 * GNU `make`
 * `pthreads`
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -114,33 +114,38 @@ Add this to your `Cargo.toml`:
...
+will be used to compile and statically link the library. The `dynamic-linking`
+feature can be used to instead dynamically link rdkafka to the system's version
+of librdkafka. Example:
 
 ```toml
-[dependencies.rdkafka]
-version = "~0.21"
-features = ["ssl", "sasl"]
+[dependencies]
+rdkafka = { version = "0.22", features = ["cmake-build", "dynamic-linking"] }
```
fede1024: Is cmake-build necessary when using dynamic linking?

benesch: D'oh, good catch, thanks!

benesch:
Thank you for building the library in the first place! Happy to keep things moving forward.

fede1024:
Should I build a new release? Or are there still open issues that we want to solve first?

benesch:
Yes, please! Folks seem to be ready for a new release, and there’s nothing else that seems urgent. (I want to get to the futures stuff ASAP, but better to put that in 0.23 instead of the next release, since there’s a lot of stuff in the ecosystem that’s still on tokio 0.1, and those folks could benefit from the past six months of compatible improvements here.)

PR #192, opened 2019-12-11
Don't create an extra thread when running message stream

dbcfd:
@benesch This gets rid of the extra thread and futures executor block_on, without introducing a tokio dependency.

### DIFF
--- a/src/consumer/stream_consumer.rs
+++ b/src/consumer/stream_consumer.rs
@@ -1,5 +1,4 @@
 //! Stream-based consumer implementation.
-use std::future::Future;
 use std::pin::Pin;
 use std::ptr;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -7,7 +6,6 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::time::Duration;
 
-use futures::channel::oneshot;
 use futures::Stream;
 use log::*;
 use pin_project::pin_project;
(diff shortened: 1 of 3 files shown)


src/consumer/stream_consumer.rs
```diff
@@ -79,65 +71,65 @@ unsafe impl Send for PolledMessagePtr {}
...
+    fn poll_next(self: Pin<&mut Self>, _cx: &mut Context) -> Poll<Option<Self::Item>> {
+        let this = self.project();
+        debug!("Polling stream for next message");
+        loop {
+            if this.should_stop.load(Ordering::Relaxed) {
+                info!("Stopping");
+                return Poll::Ready(None);
+            } else {
+                let timeout = crate::util::Timeout::After(*this.poll_interval);
+                let res = this.consumer.consumer.poll(timeout);
```
benesch: This seems quite problematic! If someone requests a large timeout they'll starve out all other futures on this core, and if they request a small timeout they'll spin hot. To put it another way, this future never returns `Poll::Pending`, and therefore is not going to play nice with other futures. My intuition suggests that the old thread-based code will be far more performant here for many workloads.

benesch:
More specifically, I'd expect this to be slightly faster on a microbenchmark with extremely high volume consumption, where there is always a message available, but potentially much slower in a lower-volume application that needs to intersperse message consumption with other async work.

src/consumer/stream_consumer.rs
redbaron: @benesch , what if we use `message_queue_nonempty_callback` to wake this stream? Then we can do `consumer.poll(0)`

Issue #191, opened 2019-12-10
Release build on windows 


GabrielBernard:
Hi, 

I tried to compile your library (version 0.22) on Windows fith the cmake_build feature. It works wonderfully on debug, but for some reasons, the release build of the rdkafka.lib goes to src/RelWithDebInfo instead of the src/Release folder, making the cargo build --release command fail with error: could not find library `rdkafka`, perhaps an -L flag is missing? 

Of course, copying the result to a src/Release folder makes everything compile, but it would be neat if the cargo:rust-link-search variable was automatically updated to search the src/RelWithDebInfo folder instead of the src/Release folder for a better user experience on Windows. Or you know, if the release version of the library was generated in the Release folder instead of the RelWithDebInfo folder. Whatever is easiest for you. 

Thank you for your time! I am not a pro on foreign function interface programming with rust so I am not sure on how to make those changes, otherwise I would make you a pull request instead of this issue.

benesch:
Gah, thanks, looks like we were handling release and debug properly, but not release + debug info. I've opened a PR (#194) to fix this properly.

benesch:
Thanks very much for the report, @GabrielBernard, and thanks especially for being so gracious! Your kind words are very much appreciated.

PR #193, opened 2019-12-11
Build in both debug and release mode in CI

benesch:
This would have caught #191.


benesch:
Never mind. Sadly this didn't actually catch the bug (you need to build using a release profile that re-enables debug information), and testing all three permutations (release, release with debug info, and debug) is prohibitively slow.

PR #196, opened 2019-12-11
Auto-generate reverse mappings for `rd_kafka_resp_err_t`

igozali:
Wanted to see if you'd be interested in this?

The idea is so that we don't have to manually update `helpers.rs`. Moving the `bindgen` logic to `build.rs` would also allow us to generate the bindings as part of the `cargo build` process.


benesch:
Hi @igozali, thanks very much for the PR. Unfortunately I just recently switched us back to manually running bindgen (https://github.com/fede1024/rust-rdkafka/pull/163/commits/1c7a58c54e66a49a8282697a62bcd089bd2c858d), as running it during compile time requires users to have libclang installed, which is problematic for more than a few folks. 

The enum stuff looks interesting! Could we use num_enum instead of rolling our own support, though?

igozali:
Oh, I see, good to know, and oops, didn't know about `num_enum` 🤦‍♂  Was quite fun learning about procedural macros, though.

Without bindgen inside `build.rs`, I wasn't able to figure out how to add an additional trait to derive from for `rd_kafka_resp_err_t` using bindgen (which we seem to need even if we use `num_enum`). Actually, even my solution right now using `str::replace` is totally a hack. Do you have any thoughts on this? Otherwise it'd be a `sed` somewhere in `update_bindings.sh` probably.

benesch:
This is great, thanks! Merged with some minor modifications to elide the now-unnecessary primitive_to_rd_kafka_resp_err_t function entirely.

benesch:
And yeah, unfortunately there is presently no way to ask bindgen to derive nonstandard traits for the types, so we're stuck with regex mashing. Oh well. There's an open bindgen issue about it, so perhaps it gets fixed eventually: https://github.com/rust-lang/rust-bindgen/issues/1097

igozali:
Sweet, also I like that perl version of regex mashing better 😃

PR #207, opened 2020-01-11
Fix doc generation on docs.rs by disabling librdkafka linking.

fede1024:


benesch:
Hi @fede1024, thanks for dealing with the problem I created! I think this would work, but there's also a slightly easier solution in my opinion, which is to remove the hack that was causing docs.rs generation to fail—see #208. Basically, now that we don't let mklove build in-tree, we don't need the CMake build to try to remove an in-tree config.h, because we'll never write that in-tree config.h in the first place.

fede1024:
Ah! Much better. No need for this change anymore then. Thanks!

PR #209, opened 2020-01-13
derive Debug for ClientConfig

prettynatty:
Debug is needed when ClientConfig is put into Context struct of some web frameworks


benesch:
Thanks, this seems worthwhile to me. We're working on getting 0.23.1 out the door, so holding off on merging until that happens.

benesch:
Thanks!

PR #213, opened 2020-01-29
Add Clone to OwnedMessage

sphw:
There doesn't seem to be a reason that `OwnedMessage` should not implement `Clone`. Certain use cases like using a `tokio::sync::broadcast` channel require `Clone` to be implemented.


benesch:
Agreed, thanks!

Issue #217, opened 2020-02-16
BaseRecord timestamp


justmike2000:
It would be wonderful if both the documentation and src comment mentioned that the BaseRecord timestamp function that takes an i64 parameter mentions that the function parameter expects milliseconds.

benesch:
Sure! Can I talk you into submitting a PR?

justmike2000:
https://github.com/fede1024/rust-rdkafka/pull/218

benesch:
Fixed in #218.

PR #218, opened 2020-02-17
Base record timestamp fix issue #217

justmike2000:
Added comments to BaseRecord timestamp to describe kafka expects timestamps in millisecond format.

https://github.com/fede1024/rust-rdkafka/issues/217


benesch:
Thanks very much! I adjusted the wording to be a bit more specific, but otherwise looked good!

PR #221, opened 2020-02-25
Adding MUSL feature for cross compiling to alpine

stevenleadbeater:
This allows cross compiling from an Ubuntu dev machine to an alpine docker image on the host dev box. I am quite new to rust and honestly feeling around in the dark on the CMake / C and C++ stuff. This will not work on windows machines, I'm happy to look in to warning about that if the flag is set somehow and improving the documentation of this feature if the approach is even palatable! The reasoning behind attempting this is that cargo is caching a lot of depdendencies on the build machine, there are approaches that use multi stage docker files to do the build, this however leaves a lot more of the CI/CD process in a docker file than I would like. If I can copy a statically linked cross compiled binary in to an alpine container, I can run tests from cargo outside of docker, have more of the build process in whatever pipeline definition I'm using, have a small portable container produced at the end and leverage cached target directories for local development and possibly even CI/CD machines. All comments welcome!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rdkafka"
-version = "0.23.1"
+version = "0.23.2"
 authors = ["Federico Giraud <giraud.federico@gmail.com>"]
 repository = "https://github.com/fede1024/rust-rdkafka"
 readme = "README.md"
@@ -11,7 +11,7 @@ categories = ["api-bindings"]
 edition = "2018"
 
 [dependencies]
-rdkafka-sys = { path = "rdkafka-sys", version = "1.3.1", default-features = false }
+rdkafka-sys = { path = "rdkafka-sys", version = "1.3.2", default-features = false }
 futures = "0.3.0"
 libc = "0.2.0"
 log = "0.4.8"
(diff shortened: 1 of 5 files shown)


benesch:
Hey @stevenleadbeater, thanks for working on this! Unfortunately I don't think that Cargo features are an appropriate way to trigger cross-compilation. For example, `musl-gcc` may not be the name of the cross-compiling GCC on every system, and might not live in `/usr/bin`. My goal instead is to make sure that the rust-rdkafka build system is powerful enough to allow users to override these values appropriately via environment variables.

I'd be happy to instead accept a PR that adds a CI job that tests the build on musl, along with documentation that points users who wish to compile against musl at the working CI job.

stevenleadbeater:
Fair enough @benesch it actually works if I prefix my build with `CC="musl-gcc -static -Os" CXX="/usr/bin/g++"`

benesch:
Awesome, glad to hear it!

benesch:
Although on a second look using a cross-compiling `musl-gcc` and a non-cross-compiling g++ seems likely to be very problematic? Like perhaps that is only working because you're not actually linking any of the C++ code in librdkafka? Who knows though; cross-compiling always requires a lot black magic and crossed fingers.

stevenleadbeater:
I know what you mean, it's not been easy to get this working. the command I used was `CC="musl-gcc -static -Os" CXX="/usr/bin/g++" cargo build --release --target=x86_64-unknown-linux-musl` from what I've gathered, this means I'm incanting magic words in to CMake to tell it to use the musl lib c aware c compiler and normal c++ one. It results in ldd telling me `not a dynamic executable` which is what I wanted and it all seems to work. I'm really not a C or C++ dude, neither am I really a proper rust person yet. I'm taking a dive in to these deep waters from a Java, JS and c# background so I'm pretty much just poking things and seeing what happens!

stevenleadbeater:
I think you're right about the feature not being the place to put this though. I knew it was off while I was doing it. I'm not 100% on Travis at all maybe 5% tops but I'll see if I can pull together a working example for cross compiling a static binaries from an ubuntu host to an alpine target and add some documentation instead. I want to help and don't want to be responsible for muddying up your repo either :)

Issue #220, opened 2020-02-23
Memory keeps growing


llitz:
Not sure if this is a problem or where it is but after creating a program with FutureProducer my memory skyrocketed to 10Gb usage by having 3 FutureProducers running at the same time (otherwise I get limited to 50k EPS).

Even when running only one memory usage was at 4gb.

After chasing this a little while, I have tried running the asynchronous_processing example, and memory also grows there at least until 1GB, any idea what can be causing this?

![ss-2020-02-23-130822](https://user-images.githubusercontent.com/22463547/75119221-07f9e480-563e-11ea-99ff-c2149ca27ba7.png)

gklijs:
I think this is just page cache from c, and will be freed when needed. I did some tests before, see https://www.confluent.io/blog/getting-started-with-rust-and-kafka/ at the memory section and memory use is amazingly low, but it's measured excluding page cache. It might also be regression through.

llitz:
I don't think that's the case, I had memory usage grow all the way up to 32gb.

For now, I am limiting consumption loop to 1m entries and ending then starting it all over. This reduces how fast I can process all messages, but memory is "limited" that way.

This only happens when future producer is involved /=

tobz:
For what it's worth, I've also noticed a growing memory usage on a consumer-based application I'm working on.  I tracked it down as being related to the buffers allocated for doing gzip decompression, so maybe this is related to message _compression_ as you produce?

llitz:
I don't think so, I have compression none and the data is not compressed in the input topic either.

gklijs:
And on the broker? Because in case of compression it might be taking the property from the broker configuration.

llitz:
The data is in clear text there.

If I disable the producer the program doesn't use more than 250mb of memory to read 200k messages per second. They are around 600 bytes each, some larger, some smaller.

Marwes:
Also seeing memory leaks in an application that contains a kafka producer. Might very well be something else causing it but the existence of this issue makes kafka the most likely candidate

benesch:
Is anyone able to get a heap profile?

Marwes:
I ran the app through valgrind and didn't find anything. No idea how to get heap profiles out of production at this point.

tobz:
I can provide a macOS Instruments memory profile; getting Valgrind to work on macOS has been a pain.  Will see if I can do that unless macOS Instruments works for you.

llitz:
[...]

PR #223, opened 2020-03-11
refactor: Remove the need for mem::forget in some cases

Marwes:
Looked into if there was an obvious fix for #220 but nothing so far. Still this improves things a bit.


benesch:
Hey @Marwes, thanks for the patch!

> Still this improves things a bit.

Wait, so your claim is that this improves memory usage a bit? That surprises me. By my read this code is functionally equivalent to the old code. 

Happy to merge this because I think this formulation is less error prone, assuming my understanding of its behavior is correct.

Marwes:
Only meant it improves the code a bit, it has no practical effect :)

benesch:
Oh good! We agree. :)

PR #233, opened 2020-04-08
docs: expose name gssapi-vendored

sploiselle:
Seems like this feature is enabled in the code as [`gssapi-vendored`](https://github.com/fede1024/rust-rdkafka/blob/master/rdkafka-sys/Cargo.toml#L57).


benesch:
Thanks! 🤦‍♂

PR #237, opened 2020-04-24
Report full windowing statistics

### LINKED ISSUE #229: Windowing statistics do not record percentile information
Rdkafka 1.3.0 reports percentile information for window statistics: https://github.com/edenhill/librdkafka/blob/v1.3.0/STATISTICS.md#window-stats

This information should be deserialized in the `Window` struct so that it is available to clients interested in it. https://github.com/fede1024/rust-rdkafka/blob/master/src/statistics.rs#L58

nate-onesignal:
rdkafka-1.3.0 supports a much richer set of windowing statistics than
this library previously exposed. Most interesting is the percentile
information, which allows users to see approximately how many requests
are falling into performance issues.

Resolves #229

### DIFF
--- a/src/statistics.rs
+++ b/src/statistics.rs
@@ -61,6 +61,15 @@ pub struct Window {
     pub avg: i64,
     pub sum: i64,
     pub cnt: i64,
+    pub stddev: i64,
+    pub hdrsize: i64,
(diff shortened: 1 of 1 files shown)


benesch:
Thank you very much!

Issue #211, opened 2020-01-15
Release 0.22 build from cargo fails after the release of 0.23


tkasu:
I'm having problems building my [project](https://github.com/tkasu/hsl-data-platform/tree/master/applications/vehicle-pos-data-forwarder) that is using release 0.22. 

It seems that problems started after the release of rust-rdkafka 0.23 and rdkafka-sys 1.3.1.

```
   Compiling rdkafka-sys v1.3.1
  ...
   Compiling rdkafka v0.22.0
error[E0425]: cannot find function `primitive_to_rd_kafka_resp_err_t` in crate `rdsys`
   --> /Users/tkasu/.cargo/registry/src/github.com-1ecc6299db9ec823/rdkafka-0.22.0/src/client.rs:411:22
    |
411 |     let err = rdsys::primitive_to_rd_kafka_resp_err_t(err)
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `rdsys`

```

Release 0.22 is fetching rdkafka-sys 1.3.1 from which the function primitive_to_rd_kafka_resp_err_t has been removed.

benesch:
That's unfortunate. Perhaps we need to be using strict version requirements in `rdkafka`. For now you should be able to work around the issue by adding a strict dependency on the correct version of rdkafka-sys to your project:

```
rdkafka = "0.22.0"
rdkafka-sys = "=1.2.2"
```

Alternatively perhaps we need to be bumping the major version of rdkafka-sys whenever we release a new version, since most versions involve semver breaking changes—but then we'd lose the parity with upstream's versioning. @fede1024 do you have thoughts?

FSMaxB-dooshop:
Why don't you use [tilde requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#tilde-requirements) to specify rust-rdkafka's dependency on rdkafka-sys?

This way you should be able to prevent minor version upgrades.

tkasu:
@benesch thanks that works! Nevertheless, if this could be handled in rdkafka's dependencies it could potentially help at least new rustaceans like me in the future=).

benesch:
> @benesch Why don't you use tilde requirements to specify rust-rdkafka's dependency on rdkafka-sys?

I'm worried that's not sufficient! I'm not sure rdkafka upstream adheres to semver.

benesch:
[...]

PR #240, opened 2020-05-02
rdkafka-sys: upgrade to librdkafka v1.4.2

benesch:
This supersedes #234, which is unmergeable due to an upstream breakage of the Windows build. That should be fixed in v1.4.2. Still awaiting the official release, but kicking things off with rc2 here.

Also allow rdkafka and rdkafka-sys to be versioned independently by
making rdkafka-sys follow SemVer. This will fix future scenarios like
the one described in issue #211.


benesch:
> Do we need to update rdkafka-sys/tests/version_check.rs as well?

Great point, thanks! Looks like that test is not getting run in CI. Addressing that now.

PR #241, opened 2020-05-06
refactor: Homogenize the unsafe around drop (and more)

Marwes:
Did a pass through the crate and cleaned up a lot of the logic around drops. The `NativePtr` were added as a small abstraction around an owned ptr and where I could, I moved the allocated ptr into this structure as early as possible to ensure drops were not forgotten.

Fixed a few other small things on the way as well.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -24,7 +24,7 @@ use rdkafka_sys::types::*;
 use crate::client::{Client, ClientContext, DefaultClientContext, NativeQueue};
 use crate::config::{ClientConfig, FromClientConfig, FromClientConfigAndContext};
 use crate::error::{IsError, KafkaError, KafkaResult};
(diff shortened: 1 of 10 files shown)


benesch:
Thanks very much for putting this together. This broadly seems great, but I'll need to take a closer look later. Did you manage to track down the memory leak you were experiencing?

Marwes:
Nope, still no idea. Were actually looking into another issue this time around where the kafka producer would be unable to send more messages. Not seeing any cause for that it this crate though, probably in librdkafka itself.

benesch:
Thanks again for all of this! Great stuff. I extracted stuff into other PRs and made some minor cosmetic nitpicks, but for the most part this landed unchanged. I squashed the `NativePtr` stuff down into one commit here that will merge shortly. 🍻

PR #248, opened 2020-05-13
Avoid unnecessary lock in threaded producer

benesch:
The JoinHandle is only accessed by the thread that holds the
ThreadedProducer, so it doesn't need to be wrapped in a RwLock. The
stop signal is already made threadsafe by use of an atomic bool.


benesch:
Inspired by #241.

PR #251, opened 2020-05-14
Require Tokio for stream consumer

benesch:
Requiring Tokio admits a much simpler implementation that does not
require a separate thread or executor. The dependency on Tokio is
optional (on by default) in order to allow folks who are using only the
BaseConsumer to avoid the Tokio dependency.

Fix #192.

### DIFF
--- a/src/consumer/stream_consumer.rs
+++ b/src/consumer/stream_consumer.rs
@@ -2,14 +2,13 @@
 
 use std::pin::Pin;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 use std::task::{Context, Poll};
-use std::thread::{self, JoinHandle};
-use std::time::Duration;
 
 use futures::channel::mpsc;
 use futures::{SinkExt, Stream, StreamExt};
-use log::{debug, trace, warn};
+use log::{debug, trace};
+use tokio::time::{self, Duration};
 
 use rdkafka_sys::types::*;
 
@@ -18,7 +17,7 @@ use crate::consumer::base_consumer::BaseConsumer;
 use crate::consumer::{Consumer, ConsumerContext, DefaultConsumerContext};
 use crate::error::{KafkaError, KafkaResult};
 use crate::message::BorrowedMessage;
-use crate::util::{NativePtr, Timeout};
+use crate::util::{NativePtr, OnDrop, Timeout};
 
 /// Default channel size for the stream consumer. The number of context switches
 /// seems to decrease exponentially as the channel size is increased, and it stabilizes when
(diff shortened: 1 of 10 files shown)


src/consumer/stream_consumer.rs
```diff
@@ -175,43 +177,25 @@ impl<C: ConsumerContext> StreamConsumer<C> {
...
-                    consumer.as_ref(),
-                    sender,
-                    should_stop.as_ref(),
-                    poll_interval,
-                    no_message_error,
-                );
-            })
-            .expect("Failed to start polling thread");
-        *self.handle.lock().unwrap() = Some(handle);
+        tokio::spawn(poll_loop(
```
redbaron, line 180: this will keep polling even if StreamConsumer is dropped, right?

FSMaxB-dooshop: No because on drop `should_stop` is set to `true` and `poll_loop` will stop looping and return.

src/consumer/stream_consumer.rs
```diff
@@ -155,15 +155,17 @@ impl<C: ConsumerContext> FromClientConfigAndContext<C> for StreamConsumer<C> {
...
     }
 }
 
 impl<C: ConsumerContext> StreamConsumer<C> {
-    /// Starts the StreamConsumer with default configuration (100ms polling interval and no
-    /// `NoMessageReceived` notifications).
+    /// Starts the StreamConsumer with default configuration (100ms polling
+    /// interval and no `NoMessageReceived` notifications).
+    ///
+    /// **Note:** this method must be called from within the context of a Tokio
```
redbaron, line 167: maybe provide

```
pub fn start_with_timer<F>(...., timerFn: F) -> MessageStream<C>
   where F: Fn() -> Future
```

And make `start_with` conditional on tokio feature?

src/consumer/stream_consumer.rs
```diff
@@ -89,35 +88,37 @@ impl<'a, C: ConsumerContext + 'a> Stream for MessageStream<'a, C> {
...
     poll_interval: Duration,
     send_none: bool,
 ) {
-    trace!("Polling thread loop started");
+    trace!("Polling task loop started");
+    let _on_drop = OnDrop(|| trace!("Polling task loop terminated"));
     while !should_stop.load(Ordering::Relaxed) {
-        trace!("Polling base consumer");
-        let future_sender = match consumer.poll_raw(Timeout::After(poll_interval)) {
+        let backoff = time::delay_for(poll_interval);
```
redbaron, line 101: this strictly speaking is not poll interval and changes semantic of this value. Previously if poll_interval was 1minuted, but data arrived in 1 second, it were to become available in 1 second,  now it is 1 minutes.  This probably not a problem in practice, but it should at least be reflected in docs.

redbaron: actually it is a problem for low latency use cases.

src/consumer/stream_consumer.rs
benesch: Yep, exactly. Actually waiting for the task to drop requires support for an `AsyncDrop` sort of thing, which doesn't exist at the moment. It's not necessary to wait, though, because the `Arc<BaseConsumer>` will keep the consumer alive as long as the task is alive.

src/consumer/stream_consumer.rs
benesch: Yeah it's never really been the poll interval; it's always been the maximum amount of time to wait for a message. Your point about the low latency use cases is well taken though. I'll address that in a follow up PR.

src/consumer/stream_consumer.rs
benesch: I don't think the `time` feature is particularly onerous to uncondtionally require. It only introduces a dependency on `slab` which is quite lightweight.

redbaron: I dont agree, it introduces dependency on tokio. Previously rust-rdkafka crate could be used with async-std or smol runtimes.

src/consumer/stream_consumer.rs
redbaron: What was the rush to merge it? You knowingly introduced regression, I don't get it.

benesch: This PR was already rather large. There are no stability guarantees on master. The problem will be fixed before the next release.

PR #256, opened 2020-05-17
Simplify stream consumer, restore support for runtimes other than tokio

redbaron:
This simplifies consumer to become a single straightforward Stream with no background tasks.
    
Mandatroy polls on set intervals are implemented using heartbeat_poll combinator, where every time
stream is Pending timer is armed, which when fired polls stream again.
    
Previous StreamConsumer::{start,start_with} are now feature gated. New generic method start_with_heartbeat is  introduced. It workes with any runtime capable of handling timers.

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -222,6 +222,10 @@ impl<C: ConsumerContext> Consumer<C> for BaseConsumer<C> {
         self
     }
 
+    fn client(&self) -> &Client<C> {
+        &self.client
+    }
+
     fn subscribe(&self, topics: &[&str]) -> KafkaResult<()> {
         let mut tpl = TopicPartitionList::new();
         for topic in topics {
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -17,15 +17,11 @@ use crate::topic_partition_list::{Offset, TopicPartitionList};
 use crate::util::{cstr_to_owned, Timeout};
 
 pub mod base_consumer;
-#[cfg(feature = "tokio")]
-#[cfg_attr(docsrs, doc(cfg(feature = "tokio")))]
 pub mod stream_consumer;
 
 // Re-exports.
 pub use self::base_consumer::BaseConsumer;
-#[cfg(feature = "tokio")]
-#[cfg_attr(docsrs, doc(cfg(feature = "tokio")))]
-pub use self::stream_consumer::{MessageStream, StreamConsumer};
+pub use self::stream_consumer::StreamConsumer;
 
 /// Rebalance information.
 #[derive(Clone, Debug)]
--- a/src/consumer/stream_consumer.rs
+++ b/src/consumer/stream_consumer.rs
@@ -1,17 +1,15 @@
 //! Stream-based consumer implementation.
 
-use std::future::Future;
 use std::pin::Pin;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll, Waker};
+use std::time::Duration;
 
-use futures::{ready, Stream};
-use log::trace;
-use tokio::time::{self, Duration, Instant};
-
-use rdkafka_sys as rdsys;
 use rdkafka_sys::types::*;
+
+use futures::stream::{self, Stream};
+use futures::{ready, Future, FutureExt};
+
+use std::sync::Mutex;
 
 use crate::client::{ClientContext, NativeClient};
 use crate::config::{ClientConfig, FromClientConfig, FromClientConfigAndContext, RDKafkaLogLevel};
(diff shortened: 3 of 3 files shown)


benesch:
Thanks, but the behavior change here makes me nervous. The current `StreamConsumer` does not need to be polled (in the `stream.next().await` sense) in order to stay part of the consumer group, while the new implementation does. That is a very subtle breaking change that may break downstream code. Unlike a change to the API, which will cause a compile-time error, or the new dependency on Tokio, which will cause a panic on startup, disconnecting from the group might not occur until after the code has been running in production for 5m (or whatever the session timeout is).

I'm open to finding a way to permit other runtimes to stand in for Tokio, and have a similar patch in the works that abstracts over `tokio::spawn` as well. In the future it might be easier if we discuss large changes like this first, so we don't duplicate effort.

For what it's worth, I am somewhat unconvinced that there are many folks using rust-rdkafka with smol or async-std. (I'd be happy to be shown otherwise, because it makes maintaining the extra complexity feel more worthwhile.) But it is my personal opinion that neither async-std nor smol have demonstrated any concrete benefits over Tokio. Perhaps you could say more about why have chosen to use smol?

redbaron:
> Thanks, but the behavior change here makes me nervous.

This brings back old behaviour where if you don't poll you got kicked out. Before refactor separate thread was pushing to a bounded queue and as soon it was blocked, polling thread was blocked too. IMHO it is correct behaviour, there is no point in pretending to be alive, when you are not.

> Perhaps you could say more about why have chosen to use smol?
It is not about one being better than the other, it is about freedom of choice.

benesch:
> It is not about one being better than the other, it is about freedom of choice.

Yes, this is exactly the point I’m driving at. Why is it important to have pluggable runtimes if the runtimes are not meaningfully different from one another? If they are all equivalent, then having three separate APIs just means three times as much work for maintainers of async/await libraries. 

I would be delighted if someone could point out legitimate tradeoffs, like “Tokio has better throughput, but smol has better latency, and depending on your Kafka workload you might see 20% better performance with smol.” But to my knowledge, no one has presented such evidence.

As I mentioned above, it is important to me that rust-rdkafka is widely useful, and evidently that means supporting multiple async runtimes. But to satisfy my own personal curiosity, I would like to know why users would prefer async-std or smol over Tokio. I know of no serious, concrete problems with Tokio; only Stjepan’s vague concerns that it is too complicated. That hasn’t matched my experience with Tokio, either.

redbaron:
> If they are all equivalent, then having three separate APIs just means three times as much work for maintainers of async/await libraries.

No, thats not 3 times as much work if core logic is independent from runtimes. Any new runtime is just few wrapper functions at the edge then. You don't even need to include support for every runtime out there, library users can do it themselves, it is zero work for you then.

Either way , this PR streamlines code, multiple runtime support comes as consequence of that. I'd like to ask you to discuss tech aspects of this PR, not broad questions of runtime differences, motivation or preferences.

From what I can see, it is not breaking anything from last released version and has fewer moving parts and doesn't lose notifications in a rare, when librdkafka makes a call just after first Pending poll, do you think it is strong enough case to merge it?

benesch:
Yes, as I mentioned in my first comment I am open to an approach like this,
but already have a patch in the works that preserves the background thread.
I will find the time soon to open a PR with that patch, hopefully later
tonight, and we can discuss.

benesch:
As promised, I've opened #260 with my patch. It takes a more targeted approach and does not require rewriting the `StreamConsumer`, but instead introduces an `AsyncRuntime` trait that has both `spawn` and `delay_for` methods that the user can implement using whatever runtime they like.

I would prefer to keep the background task so that you don't need to think about whether you are calling poll frequently enough, but if there are other pieces of this PR that you would like me to consider integrating into #260, please let me know. Thank you for catching that race condition, by the way. I believe that is also fixed in #260.

> No, thats not 3 times as much work if core logic is independent from runtimes. Any new runtime is just few wrapper functions at the edge then. You don't even need to include support for every runtime out there, library users can do it themselves, it is zero work for you then.

Coming up with the correct interface is not trivial. I spent several hours coming up with the approach in #260 that minimizes the number of backwards incompatible changes, and verifying the interface's usefulness required writing an sample `AsyncRuntime` implementation for smol.

Plus, users will inevitably implement the `AsyncRuntime` trait incorrectly, or there will be bugs in the underlying runtime, and those issues will present as issues in rust-rdkafka, which increases the maintenance burden.

> I'd like to ask you to discuss tech aspects of this PR, not broad questions of runtime differences, motivation or preferences.

Justification for introducing additional complexity is an important and technical concern. I don't think it is too much for me to ask why you prefer smol in a PR where you expose hooks so that you can plug in smol. I am legitimately curious as to why people are moving away from Tokio.

All this to say: I think supporting pluggable runtimes makes sense, but it is not without its costs, costs that will be largely shouldered by the rust-rdkafka maintainers.

benesch:
I've merged #260. Let me know if you have lingering concerns. There is an example of how to plug in smol here: https://github.com/fede1024/rust-rdkafka/blob/c0cf67d7b6579b5b8f9fc0e53905f5646cb8dd5f/examples/smol_runtime.rs#L19-L34

PR #259, opened 2020-05-17
Fix blocking behavior of FutureProducer::send

benesch:
@FSMaxB-dooshop I incorporated your fix from #222 here, along with a rewrite to use `tokio::time::delay_for` rather than `std::thread::sleep`. Let me know what you think. I added one test so far for the successful case (which was evidently totally untested before), but I'm going to try to add another test for the `QueueFull` case before merging.

----

The `block_ms` parameter of `FutureProducer::send` was previously buggy,
and did not behave at all as the documentation claimed. Make several
improvements:

  * Rename the parameter to `queue_timeout`, to better reflect its
    purpose in retrying only the queue phase (not the entire send).

  * Adjust the type from an `i64` representing milliseconds to
    `util::Timeout`, for better consistency with other APIs that allow
    configurable timeouts.

  * Use `tokio::time::delay_for` internally rather than blocking the
    thread with `std::time::sleep`, as the latter is not acceptable in
    asynchronous code.

Closes #222.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,29 @@ See also the [rdkafka-sys changelog](rdkafka-sys/changelog.md).
   The old implementation required a separate thread and a separate futures
   executor.
 
+* **Breaking change.** Overhaul the `FutureProducer::send` method. The old
+  implementation incorrectly blocked asynchronous tasks with
+  `std::thread::sleep` and the `block_ms` parameter did not behave as
+  documented.
+
+  The new implementation:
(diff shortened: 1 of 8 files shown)


benesch:
The additional test is in, as promised, and things looks good. I'm going to merge to unblock another patch that adds support for pluggable asynchronous runtimes, but do let me know if you have any concerns or feedback, @FSMaxB-dooshop!

PR #263, opened 2020-05-21
Reparallelize CI tests

benesch:
CI was flaky when running multiple tests in parallel. Re-enable to start
to track down the bug.


benesch:
I've spent a while debugging this now, and I'm fairly confident that we're just overloading Travis CI (the VM only has two cores, and needs to run Kafka, the librdkafka threads, and the Rust test threads) such that librdkafka gets stuck in a timeout loop. I no longer think this needs to block the release of 0.24.0/2.0.0+1.4.2.

PR #271, opened 2020-06-09
Add Offset::OffsetTail

athre0z:
This allows specifying negative offsets reading from the end of a partition. It mirrors what the `RD_KAFKA_OFFSET_TAIL` macro does in the C lib.

Unfortunately I wasn't able to fully run all tests because `test_topics` topic fails for me (also on master without my changes applied) with a timeout error despite me having a Kafka instance listening on `localhost:9092`.

Let me know if you want me to change anything!

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -74,6 +74,19 @@ See also the [rdkafka-sys changelog](rdkafka-sys/changelog.md).
     producers to a new `Producer` trait. This trait is analogous to the
     `Consumer` trait.
 
+* **Breaking change.** Several `TopicPartitionList`-related methods now return
+  `Result<T, KafkaError>` rather than `T`:
+
+  * `TopicPartitionListElem::set_offset`
+  * `TopicPartitionList::from_topic_map`
+  * `TopicPartitionList::add_partition_offset`
+  * `TopicPartitionList::set_all_offsets`
+
+  This was necessary to properly throw errors when an `Offset` passed to one
+  of these methods is representable in Rust but not in C.
+
+* Support end-relative offsets via `Offset::OffsetTail`.
+
 * Fix stalls when using multiple `MessageStream`s simultaneously.
 
   Thanks to [@Marwes] for discovering the issue and contributing the initial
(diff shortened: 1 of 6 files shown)


benesch:
Thanks, this looks good! Unfortunately it occurs to me that `Offset(i64)` is a fairly dangerous API to have lying around; `Offset::Offset(-2)` might look like it refers to the offset two from the end, when in fact it results in the logical `Offset::Beginning`, if I'm reading this code right. Hrm. The situation is arguably worse with `OffsetTail` because `OffsetTail(i64::MAX)` is going to explode.

Maybe it doesn't much matter, but I need to think on this for a bit.

athre0z:
Yeah, that certainly is misleading. In fact, that was the first thing I tried but due to the tail offset region being shifted by -2000, it fortunately failed instead of leaving me with the wrong amount of tail messages when I requested `-10`.

We could just add a second type `struct RawKafkaOffset(pub i64)` and `TryFrom<KafkaOffset> for RawKafkaOffset` + `TryFrom<RawKafkaOffset> for RawKafkaOffset` and make the `KafkaOffset::Offset` field unsigned?

In `RawKafkaOffset` -> `KafkaOffset` we would reject any constant that isn't in a known range.
In `KafkaOffset` -> `RawKafkaOffset` we essentially just reject the `i64::MAX..u64::MAX` range.

Problem with this approach is that if rdkafka adds more cases with magic constants, TryFrom will fail for those if we don't support all of them exhaustively.

An alternative could be just renaming what is currently known as `KafkaOffset::Offset` to `KafkaOffset::Raw` and then add `KafkaOffset::Offset` back as an unsigned version. This design would require less changes and not break for unknown raw values.

benesch:
> An alternative could be just renaming what is currently known as `KafkaOffset::Offset` to `KafkaOffset::Raw` and then add `KafkaOffset::Offset` back as an unsigned version. This design would require less changes and not break for unknown raw values.

There's still trouble there, though, because what if you show up with `Offset(u64::MAX_VALUE)`? That won't fit in an isize, and there isn't an great place to return an error in some of the APIs that take offsets. E.g., setting the offset on a TPL does not currently return an error, and it could be kind of surprising if that's a fallible operation. But maybe that is indeed the answer.

benesch:
> But maybe that is indeed the answer.

Indeed, that's the solution I went with. I've updated this PR accordingly. Sorry for the very long delay!

PR #277, opened 2020-07-09
Expose sasl-plain feature in newn version of sasl2-sys

benesch:
This permits using SASL PLAIN authentication when vendoring libsasl2.


benesch:
Never mind; looks like librdkafka ships its own SASL PLAIN implementation rather than using the one provided by libsasl2.

PR #280, opened 2020-07-28
Add KafkaError::is_fatal to determine if an error was fatal

Licenser:
this adds the ability to query `error.is_fatal()` to distinguish between fatal and non-fatal errors allowing responding differently.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -224,3 +224,23 @@ impl From<ffi::NulError> for KafkaError {
         KafkaError::Nul(err)
     }
 }
+
+impl KafkaError {
+    /// Returns if an error is `Fatal` and requires reinitialisation.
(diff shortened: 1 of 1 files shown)


benesch:
I've opened up #281 to make it possible to get at the underlying RDKafkaError more easily. I'm a bit leery of adding an `is_fatal` function because it seems like it might be confused with librdkafka's `rd_kafka_error_is_fatal` function (https://github.com/edenhill/librdkafka/blob/39796d359898c07ea422849e6d7cd34cd13ec466/src/rdkafka.h#L768-L773), which does something slightly different?

Is #281 enough of a stopgap for you?

Licenser:
#281 looks good and seems to give the same result :) thank you!

PR #287, opened 2020-08-11
add get_fatal_error, closes #279

### LINKED ISSUE #279: Fatal error?
Hi,
we're seeing a rather odd error in the logs with rust-rdkafka:

```
Message production error: Fatal (Local: Fatal error)
```

And I'm not sure where to start investigating here, any ideas what could cause such an error?

Licenser:
Add `get_fatal_error` to the `Client` and expose it to consumers and producers for easy access

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -305,6 +305,25 @@ impl<C: ClientContext> Client<C> {
         Ok(unsafe { GroupList::from_ptr(group_list_ptr) })
     }
 
+    /// Returns the first fatal error set on this client instance, or `None` if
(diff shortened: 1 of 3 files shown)


benesch:
Thanks very much! I tweaked the API a bit and added a test. I'd like to limit the surface area of the method to just `Client` to start, though if we start building out support for more idempotent producer operations I'm definitely open to exposing the method on the producer too.

Licenser:
Thank you for merging :D and sure limiting the API makes sense!

PR #291, opened 2020-08-18
feat: Add a recv method on StreamConsumer

Marwes:
The borrow required by `MessageStream` makes the API awkward in some
cases. With this addition it becomes possible to just pass around the
`StreamConsumer` instead and just call `recv` instead of going through
`MessageStream`

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -139,7 +139,7 @@ impl<C: ConsumerContext> BaseConsumer<C> {
 
     /// Polls the consumer for messages and returns a pointer to the native rdkafka-sys struct.
     /// This method is for internal use only. Use poll instead.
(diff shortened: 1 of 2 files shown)


Marwes:
ping

Marwes:
pong

benesch:
Thanks very much for this, and sorry for the delay. I made a bunch of breaking changes to the `StreamConsumer` API that made this easier for me to reason about in #315, then added this method under the name `next` rather than `recv` to mirror the name of the method in the de facto standard `StreamExt` trait.

Marwes:
I actually explicitly changed from `next` to `recv` because it is close to it, but not actually `StreamExt::next`. `tokio` does the same for it's channels for instance https://docs.rs/tokio/1.0.1/tokio/sync/mpsc/struct.Receiver.html#method.recv

benesch:
Ok, I'm convinced. Renamed to `recv`.

Issue #289, opened 2020-08-13
Transaction support


roignpar:
Now that librdkafka [supports transactions](https://github.com/edenhill/librdkafka/issues/1308#issuecomment-608085254), are there any plans to make them available in rust-rdkafka as well?

I need to use transactions in a read-process-write scenario and would be happy to help with the implementation.

benesch:
Yes, I'd love to see this! I haven't thought at all about what the API would need to look like. The right place to start is probably a proposal for/prototype of the API.

roignpar:
Looking at [librdkafka] and some of its wrappers for other languages ([python], [go], [.net]) I think the way to go would be to add:

* These methods to `BaseProducer`, `ThreadedProducer` and `FutureProducer`:
```rust
fn init_transactions<T: Into<Timeout>>(&self, timeout: T) -> KafkaResult<()> {}
fn begin_transaction(&self) -> KafkaResult<()> {}
fn send_offsets_to_transaction<T: Into<Timeout>>(
    &self,
    offsets: &TopicPartitionList,
    metadata: &ConsumerGroupMetadata,
    timeout: T) -> KafkaResult<()> {}
fn commit_transaction<T: Into<Timeout>>(&self, timeout: T) -> KafkaResult<()> {}
fn abort_transaction<T: Into<Timeout>>(&self, timeout: T) -> KafkaResult<()> {}
```

Maybe it would make sense to make some of these `async` for the `FutureProducer`.

* `ConsumerGroupMetadata` struct;

* `Consumer` and `BaseConsumer` method to get consumer group metadata:
```rust
fn group_metadata(&self) -> ConsumerGroupMetadata {}
```

* Maybe another `KafkaError` variant: `Transaction`;

* more things that I may not be aware of at this time.

What do you think?


[librdkafka]: https://docs.confluent.io/current/clients/librdkafka/rdkafka_8h.html#abf3e48791cf53ac2c7e30a17ce896f74
[python]: https://docs.confluent.io/current/clients/confluent-kafka-python/index.html#pythonclient-producer
[go]: https://docs.confluent.io/current/clients/confluent-kafka-go/index.html#Producer
[.net]: https://docs.confluent.io/5.5.0/clients/confluent-kafka-dotnet/api/Confluent.Kafka.IProducer-2.html

benesch:
[...]

Issue #273, opened 2020-06-26
Event interface


vojtechkral:
Hi,
   we're using rust-rdkafka at @braiins . However, we found out that we probably wouldn't like to use the interface as is because of the requirement to spin up threads per stream and/or polling. We decided to fork it internally and switch to the event interface of librdkafka, which seems to be pretty efficient and goes quite well with the Rust's Futures model. At least a lot better compared to the callback interface, which is kind of hard to hook up with Futures/Streams without additional threads and/or long polling. I have written support for the evented interface in rust-rdkafka for our internal fork and after that we observed a multiple-times improvement in performance.

I noticed there is some work being done to get rid of the threads, however, I didn't study it in detail and if I'm looking right the evented interface is not being used...

I wanted to ask:
- Is there a technical reason why the evented interface is not used in rust-rdkafka? Is there some kind of a downside I don't see?
- Would you be interested in accepting code from us? I would probably need to do some cleanups, fill in documentation etc. for public usage, however, I don't want to spend time on that unless there is interest in it :-)

Thank you and thanks for making rust-rdkafka,
Vojtech

fede1024:
Hello!

I haven't worked on the library for a while (@benesch has been doing all the work for the one 1+ years!), but I remember that moving to the event-based librdkafka interface was one of the goals I had in mind and I was actively experimenting on ~2 years ago, so I'd definitely be interested in the change you're proposing. I can't think of any downside, but again, my mental model of the library is a bit stale.

I imagine that the rust-rdkafka public interface would remain the same?

I leave it to @benesch to comment whether he'd have time to review this change or not, since it's probably going to be a significant amount of work.

benesch:
Apologies for the delayed response—things have been busy at work lately!

I can't promise that I'll have time to review any large changes in the short term, but hopefully in the medium-to-long term. I'm definitely interested in anything that can either simplify the implementation of rust-rdkafka or improve its performance.

That said, I'm not too familiar with librdkafka's event API. To make sure we're on the same page: you're talking about `rd_kafka_conf_set_background_event_cb` and friends, right? I'm not totally sure how that would fit in to the existing `StreamConsumer` and `FutureProducer`—both of those have fairly efficient implementations now, and the `StreamConsumer` doesn't even need a separate thread. But I can definitely believe that there are simplifications and performance improvements to be had.

[...]

PR #293, opened 2020-08-20
Add producer transaction support

roignpar:
To do:
- [x] ThreadedProducer
- [ ] FutureProducer
- [x] [proper error handling](https://github.com/fede1024/rust-rdkafka/issues/289#issuecomment-677759157)
- [x] understand if it makes sense to serialize and deserialize consumer group metadata like other librdkafka wrappers do
- [x] documentation
- [x] tests


roignpar:
Opened this PR following https://github.com/fede1024/rust-rdkafka/issues/289.

roignpar:
@benesch any clue how to make the transaction methods `async` on the `FutureProducer`? librdkafka doesn't support callbacks for transaction operations.

benesch:
Sorry, I've been swamped lately. About to head off on vacation, which means I should actually have a few spare hours to look at this. But I gave it a quick skim and everything looks very promising!

> @benesch any clue how to make the transaction methods async on the FutureProducer? librdkafka doesn't support callbacks for transaction operations.

Nope, afraid not. Maybe there are some tricks along the lines mentioned in #273, but any solution along those lines wouldn't be plug n play.

roignpar:
Even with the event interface, there seem to be no transaction events defined in librdkafka.
I propose that as part of this PR the `FutureProducer` transaction methods be sync and an async API be implemented separately when it becomes feasible.
Since the methods are the same on all producer types and both `ThreadedProducer` and `FutureProducer` delegate to the base producer, would it make sense to create a `TransactionalProducer` trait similar to the `Consumer` trait?

roignpar:
While writing the tests I ran into something that I was able to reproduce using the go librdkafka wrapper and opened an [issue](https://github.com/confluentinc/confluent-kafka-go/issues/525).
Added an example and more documentation.
Feel free to change anything!

benesch:
Sorry for the massive delay here. I'm working through this today one commit at a time.

damoon:
I am very new to Rust, but I would love to see exactly once semantics for Kafka.
Anything I can do to help?

benesch:
Thanks very much for this. This was an absolutely awesome PR. I'm sorry that my review stalled out on this for so long. I took this and made a few changes in #323. The only real substantive change was fixing a few memory safety issues with `RDKafkaError`. 

By the way, there's now a `Producer` trait, to mirror the `Consumer` trait, so that's where I moved the transaction-related methods to. It's a shame that they're blocking methods on the `FutureProducer`, but there's not much we can do about that.

PR #300, opened 2020-11-04
Update to newer smol runtime

Licenser:
update examples to newer smol version


benesch:
Thanks! Wanted to avoid the dependency on `async-std`, so refactored to use the new `smol::block_on`.

PR #303, opened 2020-11-27
feat: Allow concurrent use of StreamConsumer

Marwes:
Before, each concurrent consumer would overwrite the earlier set waker
so only one task at a time would be awoken. By changing this to a Slab
where each `MessageStream` has its own slot, which is recycled on drop,
we can start multiple `MessageStream`s which concurrently polls kafka
and all get woken up when new messages arrive.

Based on #291 since these conflict (unfortunately, the recv method added there now needs to lock the context at least three times per poll, didn't see a good way to correct that without breaking changes however).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ log = "0.4.8"
 serde = { version = "1.0.0", features = ["derive"] }
 serde_derive = "1.0.0"
 serde_json = "1.0.0"
+slab = "0.4"
 tokio = { version = "0.3", features = ["rt", "time"], optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


ramn:
@benesch Hello Nikhil, I wonder if I could trouble you with reviewing this pull request?
We have been running a fork of rust-rdkafka with this patch applied for some weeks now with good results. Would be awesome to get this upstreamed!

benesch:
Hey, sorry, life's been real busy lately. I hope to make some progress on the growing backlog this weekend.

benesch:
Ok, I've finally paged most of this context back in. The `StreamConsumer::start` method long predates me, but I believe the intent was that you would only call it _once_ per `StreamConsumer`. Obviously that restriction is not enforced by the type system, though.

It seems reasonable enough to want to have multiple live `MessageStream`s for the same `StreamConsumer`, considering the underlying C object is entirely threadsafe, so I'm 👍🏽 on this. Thanks very much! I made this change independent of the `recv` addition and submitted it in #311.

PR #310, opened 2021-01-02
Vastly simplify StreamConsumer API

benesch:
As described in the changelog entry. This is an alternative to #291 that
involves breaking changes.


benesch:
Eh, I don't actually like how this turns out. Implementing `Stream` for `&StreamConsumer` is awkward. If Rust had GATs then we could implement `Stream` for `StreamConsumer` directly, but, alas, it does not.

PR #328, opened 2021-01-05
Fix references to dynamic_linking feature name

bugadani:
The correct name of the feature flag is `dynamic_linking`, not `dynamic-linking`. This PR replaces the incorrect names to the correct one.


benesch:
Whoops, thanks for catching this! Would actually prefer to go in the other direction, though, where we make the feature `dynamic-linking` exist and deprecate the old name. I've opened #331 for this.

bugadani:
That works as well, I was just surprised the docs didn't work out of the box ;) I'll update ASAP

benesch:
Oh, sorry for the crossed signals! Thanks very much!

PR #333, opened 2021-01-07
make `get` public to allow reading back config

elindsey:
It would be very useful in some cases to read back params from the final config, otherwise I end up needing to record what's been set in some piece of external state.


benesch:
Good idea! I took this and ran with it a bit to make it possible to also get librdkafka's default values for parameters and to remove parameters.

PR #338, opened 2021-01-23
use global environment variables in build.rs

fanatid:
Currently, `CFLAGS`/`LDFLAGS` respected only if `DEP_OPENSSL_VENDORED` is not defined, otherwise these variables completely rewritten.
https://github.com/fede1024/rust-rdkafka/blob/6ac37feb6931332422d604ed80a904caec1eabd7/rdkafka-sys/build.rs#L41-L45

### DIFF
--- a/rdkafka-sys/build.rs
+++ b/rdkafka-sys/build.rs
@@ -37,11 +37,21 @@ where
     }
 }
 
+fn add_to_env<K: AsRef<OsStr>>(key: K, value: &str) {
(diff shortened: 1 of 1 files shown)


benesch:
Thanks! That entire code block looked stray to me, so I just removed it. I think it was the result of a bad rebase, because both the CMake and non-CMake code paths were already handling vendored OpenSSL check separately from that block. Let me know if the fix in #339 doesn't wind up working for you.

PR #342, opened 2021-01-26
Improve `set` method of `ClientConfig`

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>


benesch:
Seems like a good idea. Thanks very much.

PR #351, opened 2021-03-09
Add async split_partition_queue to StreamConsumer

nemosupremo:
I currently need a way to consume partitions concurrently in an asynchronous context. `BaseConsumer` currently provides `split_partition_queue` for "synchronous" polling, but `StreamConsumer` provides no function. I decided to take a stab at implementing it, as it seemed simple enough; however this PR is more in the design stage.

What I opted to do was "overload" `MessageStream` to simply rollover an enum of values, either a `StreamConsumer` or `Arc<BaseConsumer>` + `Partition`.  To achieve this I also had `StreamConsumer` always hold on to an `Arc` of BaseConsumer, which I think should have a very minimal cost compared to the owned version.

The only problem I see is that each `PartitionQueue` has its lifetime tied to the `StreamConsumer` despite each `PartitionQueue` holding an Arc to base consumer. I'm not sure yet if this is an onerous restriction, however an alternate design would require a new `MessageStream` type (which I think may duplicate more code), or allowing `MessageStream` to hold an Arc to `StreamConsumer` or making the implantation of `MessageStream` generic over the "consumer".

### DIFF
--- a/examples/partition_consumer.rs
+++ b/examples/partition_consumer.rs
@@ -1,0 +1,173 @@
+use std::sync::Arc;
+
+use clap::{App, Arg};
+use futures::stream::StreamExt;
+use log::{info, warn};
+
+use rdkafka::client::ClientContext;
+use rdkafka::config::{ClientConfig, RDKafkaLogLevel};
+use rdkafka::consumer::stream_consumer::StreamConsumer;
+use rdkafka::consumer::{CommitMode, Consumer, ConsumerContext, Rebalance};
+use rdkafka::error::KafkaResult;
+use rdkafka::message::{Headers, Message};
+use rdkafka::topic_partition_list::TopicPartitionList;
+use rdkafka::util::get_rdkafka_version;
+
+use crate::example_utils::setup_logger;
+
+mod example_utils;
+
+// A context can be used to change the behavior of producers and consumers by adding callbacks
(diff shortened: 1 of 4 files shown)


nemosupremo:
Upon further research, it seems like for this to work there would need to be a type of `BorrowedMessage` that has a `'static` lifetime but also holds a reference to its consumer in an Arc.

nemosupremo:
Now the only remaining problem is much of the `Consumer` API requires a `BorrowedMessage`. A path forward would be to change the api to accept something like `AsRef<BorrowedMessage>`

nemosupremo:
One final change: I've opted to remove the boxing of `BaseConsumer` in `StreamConsumer`. It seemed silly to have the double Arc requirement.

benesch:
Thanks very much for working on this, @nemosupremo! I'm sorry I haven't had a chance to look things over yet. Things are busy at work right now, and this partition queue stuff is really subtle. Definitely a useful change though. Hopefully have time for a proper review soon.

benesch:
Sorry for the very long delay here. I put together a slightly different implementation in #411, but one that was very much based on this one! The big difference is a refactor that allowed the removal of the `MessageStreamConsumer` type.

benesch:
Thanks again for the PR!

PR #354, opened 2021-03-14
Allow access into inner context in StreamConsumerContext

marcelo140:
I was creating a custom `ConsumerContext` to customize the rebalancing logic when I noticed that `StreamConsumer` wraps up my context without providing access to it.


benesch:
Good idea! Thanks!

PR #357, opened 2021-03-23
Allow offset 0 in `Offset::to_raw`

roignpar:
Offsets usually start from 0 in Kafka topic partitions.


benesch:
Thanks very much! CI is broken in three different ways, so I'm going to incorporate this into #361 and merge there.

PR #362, opened 2021-03-30
Avoid for-loops. Prefer idiomatic Rust

hbina:
Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>


benesch:
Eh, for loops are perfectly idiomatic Rust. Personally I find the `.collect<Result<_, _>>()?` idiom to be harder to understand than a simple for loop. I'm not eager to change things just for the sake of changing them. If you have specific concerns with a specific for loop (e.g., if the iterator is more performant, and that actually matters for your use case), then we can discuss, but otherwise I'd prefer to leave things as is.

Issue #366, opened 2021-05-18
[Suggestion] Receive option as hashmap


djKooks:
Hello,
While using the module, there are a situation to receive configs as HashMap.
But in the example I could only find setting config one by one using `ClientConfig::new().set().set()...`

Is there some option to put option as hashmap such as
```
let config = HashMap::new();
// ...
let client_config = ClientConfig::new().set_config_map(config);
...
```
?

benesch:
Fixed in #367. Thanks, @djKooks!

PR #367, opened 2021-05-27
Add option to setup config as map

djKooks:
Hello,
This is update of issue https://github.com/fede1024/rust-rdkafka/issues/366.

Please review 🙏


benesch:
Thanks! I'd rather this not be so special purpose, so amended to impl `Extend` and `FromIterator` on `ClientConfig` instead.

PR #374, opened 2021-06-30
rdkafka-sys: upgrade to librdkafka v1.7.0

benesch:


benesch:
Ah yeah, the status is that CI has been broken for months and I haven't had time to look into why. I'll spend a bit on this now, one sec.

PR #388, opened 2021-09-06
introduce `ResultVec` type

crepererum:
While `Vec<Result<T, E>>` is techincally the correct type, it misses a
very important aspect of `Result<T, E>`: `must_use`. This lint ensures
that the result is actually used (aka that the user checks for errors).

By introducing a thin wrapper type, we ca get the same behavior for
operatios that return a list of results, e.g. "create topics". This
makes the API safer to use.


benesch:
Hey, thanks for the PR! Unfortunately this is a pretty breaking change, and I'm not sure it's worth it just for the addition of a `#[must_use]` annotation.

PR #395, opened 2021-09-30
Add streaming support for PartitionQueue

davidblewett:
* Add split_partition_stream method as a pair of split_partition_queue
* Add PartitionStream struct as pair of MessageStream
* Implement Stream for PartitionStream

Co-authored-by: @duarten

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -609,15 +609,15 @@ pub struct PartitionQueue<C>
 where
     C: ConsumerContext,
 {
-    consumer: Arc<BaseConsumer<C>>,
-    queue: NativeQueue,
+    pub(crate) consumer: Arc<BaseConsumer<C>>,
+    pub(crate) queue: NativeQueue,
 }
 
 impl<C> PartitionQueue<C>
 where
     C: ConsumerContext,
 {
-    fn new(consumer: Arc<BaseConsumer<C>>, queue: NativeQueue) -> Self {
+    pub(crate) fn new(consumer: Arc<BaseConsumer<C>>, queue: NativeQueue) -> Self {
         PartitionQueue { consumer, queue }
     }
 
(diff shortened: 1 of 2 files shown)


duarten (approved):

davidblewett:
@benesch any thoughts on this? I've done some testing with it in our environments and it appears to work well.

benesch:
Ok, I finally got time to look at this! So, I pushed up a reworked implementation in #411 that's a combination of your implementation, the one in #351, and a few changes of my own. The highlights:

  * `StreamConsumer::split_partition_queue` now returns a `StreamPartitionQueue` that more exactly mirrors the `PartitionQueue` type. To get messages out of a `StreamPartitionQueue`, you need to construct a `MessageQueue` via the `stream` method or use the `StreamPartitionQueue::recv` convenience method, which exactly mirrors the `StreamConsumer`. This is a bit irritating in the common case, but it avoids painful lifetime issues when you need to store the `PartitionQueue` somewhere. 
  *  The `MessageStream` type is now used for both `StreamConsumer` and `StreamPartitionQueue`, which results in a lot of code sharing.
  * Wakers are now scoped to a single `StreamPartitionQueue`. That means that messages arriving on one partition queue won't cause tasks waiting on other partition queues to get needlessly woken up. 

I'll get #411 on master and into a release before too long. Sorry for the slow review. It's a bit of a low budget operation over here. :)

By the way, I noticed there's a few of you from @datadog that have been contributing lately. First of all, thanks for sending stuff back upstream! It's super appreciated. Second, if it ever makes sense, it might be time to find a more stable maintainership situation for rust-rdkafka. I sort of accidentally fell into maintainership after @fede1024 got busy a few years ago, and I occasionally spend my 20% time @MaterializeInc looking after things, but my time is in increasingly short supply. If you or anyone else at @datadog is ever interested in pitching in more seriously, please let me know!

PR #401, opened 2021-10-29
add Debug trait to DefaultClientContext/DefaultConsumerContext

DXist:
This PR adds Debug trait derives to default client and consumer contexts.

My context is following - I've created a custom consumer context that propagates the statistics provided by librdkafka via `tokio::sync::watch` channel. The code that uses consumers either with custom or default context has a data structure with a field parameterised by context. This data structure is logged with debug formatting. This requires Debug trait implemented for both contexts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1123,7 +1123,7 @@ dependencies = [
 
 [[package]]
 name = "rdkafka"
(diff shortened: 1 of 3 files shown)


benesch:
Thanks!

PR #404, opened 2021-11-02
Add Default trait to Statistics

scanterog:
rdkafka publishes stats every configurable ms. Some of these properties are ever increasing and requires us to calculate the difference between the current value and the previous value (for example the 'tx' and 'txmsgs' fields). Setting a Default trait allow us to use the default values for fields of the struct which are required on startup (if you're storing a default Stats obj).


benesch:
Thanks, I added the `Default` to all the stats structs for good measure and will merge in #410.

PR #407, opened 2021-11-25
Create empty topic partition list for callback if needed

### LINKED ISSUE #406: Topic partition list pointer can return none
When upgrading our Kafka cluster we saw this error happen:

```
'called `Option::unwrap()` on a `None` value' in in /root/.cargo/registry/src/github.com-1ecc6299db9ec823/rdkafka-0.27.0/src/topic_partition_list.rs:201
```

It looks like taking the the pointer to a topic partition list can fail, this happens [here](https://github.com/fede1024/rust-rdkafka/blob/master/src/topic_partition_list.rs#L201). Not sure what do here, should there be some error handling in place?

thijsc:
We've seen a case where the offsets pointer in the commit callback was null. Check for this state and pass an empty topic partiton list into the callback if it happens.

Fixes #406

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -39,10 +39,16 @@ pub(crate) unsafe extern "C" fn native_commit_cb<C: ConsumerContext>(
     } else {
         Ok(())
     };
-    let tpl = TopicPartitionList::from_ptr(offsets);
+    let (forget_tpl, tpl) = if offsets.is_null() {
+        (false, TopicPartitionList::new())
+    } else {
+        (true, TopicPartitionList::from_ptr(offsets))
+    };
     context.commit_callback(commit_error, &tpl);
 
-    mem::forget(tpl); // Do not free offsets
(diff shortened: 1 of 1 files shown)


src/consumer/base_consumer.rs
```diff
@@ -39,7 +39,11 @@ pub(crate) unsafe extern "C" fn native_commit_cb<C: ConsumerContext>(
     } else {
         Ok(())
     };
-    let tpl = TopicPartitionList::from_ptr(offsets);
+    let tpl = if offsets.is_null() {
+        TopicPartitionList::new()
```
benesch, line 43: Thanks for tracking this down. This looks like a memory leak though thanks to the `mem::forget` call on L49?

thijsc: Good point. I've added some logic to handle both cases.

benesch (approved):

benesch:
Thanks! LGTM.

PR #417, opened 2021-12-30
Fix handling of NULL pointer in `NativeClient::rebalance_protocol`

### LINKED ISSUE #416: Segfault in `NativeClient::rebalance_protocol`
We are seeing a sporadic segfault when handling rebalance events in our kafka client. I have traced this down to a null pointer being passed to `CStr::from_ptr`. It appears that `rdsys::rd_kafka_rebalance_protocol` can return a null pointer in some circumstances and the Rust library doesn't handle that. I have worked around it by returning `RebalanceProtocol::None` when this happens, but I don't know if there is a better solution or a deeper underlying problem.

bruceg:
The underlying function call `rdsys::rd_kafka_rebalance_protocol` can
return a NULL pointer in the case of an error. The callers within
librdkafka treat this similarly to it returning "NONE", so we will
return `RebalanceProtocol::None` to emulate this behavior.

Closes #416

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -155,12 +155,17 @@ impl NativeClient {
     }
 
     pub(crate) fn rebalance_protocol(&self) -> RebalanceProtocol {
-        let protocol = unsafe { CStr::from_ptr(rdsys::rd_kafka_rebalance_protocol(self.ptr())) };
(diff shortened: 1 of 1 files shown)


benesch (approved):
Thanks very much!

PR #430, opened 2022-01-06
ci: test minimal versions

benesch:
Add a test that uses Cargo's unstable minimal-versions resolver to
ensure that the dependency specifactions in the Cargo.toml are correct.
It's easy to declare a dependency on v1.0 and then use a feature that's
only available in v1.1.

Also, tweak the dependabot configuration to ignore patch and minor
bumps. As a library, it's important to be as compatible as possible.
That means that for each dependency we want to use the oldest version
that is SemVer compatible with that dependency's latest release.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,9 @@
 version: 2
 updates:
-  - package-ecosystem: "cargo"
-    directory: "/"
-    schedule:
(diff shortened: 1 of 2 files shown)


benesch:
I'm going to merge this to cut down on the dependabot noise!

PR #432, opened 2022-01-12
Reduce futures dependency

robjtede:
this takes the dep tree size from 41 to 37

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -465,28 +465,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a06f77d526c1a601b7c4cdd98f54b5eaabffc14d5f2f0296febdc7f357c6d3ba"
 
 [[package]]
-name = "futures"
-version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e05b85ec287aac0dc34db7d4a569323df697f9c55b99b15d6b4ef8cde49f613"
-dependencies = [
- "futures-channel",
- "futures-core",
- "futures-executor",
- "futures-io",
- "futures-sink",
- "futures-task",
- "futures-util",
-]
-
-[[package]]
 name = "futures-channel"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f366ad74c28cca6ba456d95e6422883cfb4b252a83bed929c83abfdbbf2967d5"
 dependencies = [
  "futures-core",
- "futures-sink",
 ]
 
 [[package]]
@@ -540,12 +524,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "futures-sink"
-version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f2032893cb734c7a05d85ce0cc8b8c4075278e93b24b66f9de99d6eb0fa8acc"
-
-[[package]]
 name = "futures-task"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


benesch:
Hm, ok. ISTM this only removes `futures` and `futures-sink` from the dep tree. I'm not entirely sold on that being worth it.

robjtede:
I wanted to keep the PR small but I think more improvements can be made on top of this. Eg: futures-executor being feature exclude-able and trimming out proc macros.

benesch:
What is the goal? Are these dependencies causing problems in practice? I haven't seen any real-world Rust projects manage to avoid a proc-macro2 dependency; I don't see how you'd avoid it here, either, since it seems like futures-util depends on futures-macro which depends on proc-macro2.

robjtede:
> Are these dependencies causing problems in practice?
No

robjtede:
> What is the goal?

To make compiling a Rust project faster. Part of the issue is that the vibrant ecosystem of crates makes it easy to include too much and, in my opinion, the popular crates should endeavor to reduce dependencies for speed and security reasons.

> Are these dependencies causing problems in practice?

No, certainly not. But it feels like a self-fulfilling prophecy to say that a project can't be used without syn/quote/proc-macro2.

> ISTM this only removes futures and futures-sink

Heres the complete tree diff:
```diff
-futures v0.3.5
-futures-sink v0.3.5
-futures-io v0.3.5
-memchr v2.3.3
```

---

In summary, _this_ PR strictly reduces dependencies without seeming to add any maintenance burden to the maintainers.

robjtede:
Follow up to the further improvements mentioned, here's a small patch that removes 3 more deps:

```diff
-futures-macro v0.3.5 (proc-macro)
-proc-macro-hack v0.5.15 (proc-macro)
-proc-macro-nested v0.1.4
```

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 0a5e2e3c..49e5a28e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ members = ["rdkafka-sys"]
 rdkafka-sys = { path = "rdkafka-sys", version = "4.2.0", default-features = false }
 futures-channel = "0.3.0"
 futures-executor = "0.3.0"
-futures-util = "0.3.0"
+futures-util = { version = "0.3.0", default-features = false, features = ["std"] }
 libc = "0.2.0"
 log = "0.4.8"
 serde = { version = "1.0.0", features = ["derive"] }
diff --git a/src/consumer/stream_consumer.rs b/src/consumer/stream_consumer.rs
index 6edef850..53f78a7f 100644
--- a/src/consumer/stream_consumer.rs
+++ b/src/consumer/stream_consumer.rs
@@ -10,8 +10,8 @@ use std::task::{Context, Poll, Waker};
 use std::time::Duration;
 
 use futures_channel::oneshot;
-use futures_util::future::FutureExt;
-use futures_util::select;
+use futures_util::future::{self, FutureExt};
+use futures_util::pin_mut;
 use futures_util::stream::{Stream, StreamExt};
 use log::trace;
 use slab::Slab;
@@ -227,10 +227,13 @@ where
             async move {
                 trace!("Starting stream consumer wake loop: 0x{:x}", native_ptr);
                 loop {
-                    select! {
-                        _ = R::delay_for(poll_interval / 2).fuse() => wakers.wake_all(),
-                        _ = shutdown_tripwire => break,
-                    }
+                    let delay = R::delay_for(poll_interval / 2).fuse();
+                    pin_mut!(delay);
+
+                    match future::select(&mut delay, &mut shutdown_tripwire).await {
+                        future::Either::Left(_) => wakers.wake_all(),
+                        future::Either::Right(_) => break,
+                    };
                 }
                 trace!("Shut down stream consumer wake loop: 0x{:x}", native_ptr);
             }

```

benesch:
I reworked this a bit in #435. I managed to lose the `futures-executor` dependency unless you need the `NaiveRuntime`. I also backed out the changes to tests and examples; I'm much more comfortable with the increased maintenance overhead here when the blast radius is limited to stream_consumer.rs and future_producer.rs, rather than all the tests and examples.

PR #434, opened 2022-01-21
[producer] pass through error when calling flush

cjubb39:
Per the [rdkafka documentation](https://docs.confluent.io/5.5.1/clients/librdkafka/rdkafka_8h.html#aaff06c4372bce917c17f3c1a5d8b205d), `flush` can return `RD_KAFKA_RESP_ERR__TIMED_OUT`.  Pass that through on the `Producer` trait instead of swallowing it.


benesch (approved):
This looks great. Thanks very much!

PR #435, opened 2022-01-23
Reduce futures dependency

benesch:
In the main library, swap the dependency on `futures` for individual
dependencies on `futures-channel`, `futures-util`, and
`futures-executor`. This reduces the dependency load on downstream
projects. Additionally disable the `futures-util::select` macro and use
the slightly more cumbersome function to lose a dependency on
proc-macro-hack and proc-macro nested.

This has two changes from #432:

  * The `futures-executor` dependency is optionalized on the new
    `naive-runtime` feature.

  * The tests and examples continue to use the main `futures` crate,
    since there's no need to be so careful with dependencies in
    tests/examples. In examples particularly I think it's much clearer
    to refer to the main `futures` crate rather than its subcrates.

Closes #432.

Co-authored-by: Rob Ede <robjtede@icloud.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -466,9 +466,9 @@ checksum = "a06f77d526c1a601b7c4cdd98f54b5eaabffc14d5f2f0296febdc7f357c6d3ba"
 
 [[package]]
 name = "futures"
-version = "0.3.5"
+version = "0.3.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


benesch:
@robjtede please take a look at this when you have the chance. I'll probably merge tomorrow if I don't hear from you, but happy to take post-merge feedback.

robjtede (approved):
LGTM. +1 for the new feature flag.

PR #438, opened 2022-01-27
Fix misleading error message on pkg_config failure

qbx2:
pkg_config could fail for various reasons including dependency issues to other libs, not only when librdkafka is not found. Actually, I needed to `brew install openssl` to make it successful. By printing error together, it should be easier to troubleshoot such errors.


benesch (approved):
Makes sense, thanks!

qbx2:
I fixed the lint error, could you run the workflow again please?

PR #440, opened 2022-02-04
add `tracing` feature

guswynn:
Much of the rust ecosystem uses `tracing` over `log`

This branch adds an optional feature `tracing` to `rdkafka` that uses `tracing` log lines instead of `log` ones. See https://github.com/MaterializeInc/materialize/blob/883710649f95d232ead5872518969a15c1def542/src/kafka-util/src/client.rs#L28-L47 for an example of how people have to work around this, over using the `DefaultClientContext`

Some notes:
- `tracing::enabled!` is slightly different than `log::log_enabled!` in that it doesn't query the tracing subscriber for specifically `Event` type's. In practice, I have found this to be a vanishingly small difference, but with https://github.com/tokio-rs/tracing/pull/1900 this pr (after its landed, backported, and released), it could be possible to slightly improve
- `tracing` log levels are consts, not variants, so I have to do some re-declaration to make them importable to make the `enabled!` callsites as easy as possible to use
- I am not sure how to test this, other than just `cargo test/check`, but I need help setting up ci to exercise this

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,9 @@ jobs:
         - os: macos-11.0
         - os: windows-2019
           features: cmake-build,libz-static
+          rdkafka-sys-features: cmake-build,libz-static
+        - os: ubuntu-20.04
+          features: tracing
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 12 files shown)


guswynn:
cc @benesch

benesch:
Thanks, @guswynn! I wired up CI to at least run `cargo check` with the new `tracing` feature enabled.

Issue #415, opened 2021-12-06
support for OAuth


DonghunLouisLee:
It would be great if there is a sasl(oauthbearer) mechanism similart to https://github.com/edenhill/librdkafka/pull/2189 this issue in librdkafka.

jsurany-bloomberg:
Here's a PR https://github.com/fede1024/rust-rdkafka/pull/442

jzampieron:
Related issues: https://github.com/fede1024/rust-rdkafka/issues/757

jzampieron:
> Just scanning through open issues to pick up, but looks like this can be closed, right? PR is merged

I wouldn't. The PR, while merged, is incomplete and will bite downstream users. It's only implemented for `FutureProducer` and not `StreamConsumer` and `AdminClient`, which would be the complete client set.

PR #442, opened 2022-02-07
add support for OAuth

jsurany-bloomberg:
This PR is meant to provide the user a way to build a Client that uses the `oauthbearer_refresh_token_cb` functionality in rdkafka-sys. It also provides the plumbing  to define their own callback through the `ClientContext` trait.

- `ClientConfig` has a `pub(crate)` flag that tells the `Client` to set the callback function to that of the given `<impl>ClientContext` type.
- Default implementation of the callback just logs a message at the error level, the user needs to create their own callback function.

Example
```rust
// Define a type that overrides the default refresh_oauth_token method
struct RefreshContext {
    // relevant fields
}

impl RefreshContext {
    // other impl details
}

impl ClientContext for RefreshContext {
    fn refresh_oauth_token(&self, client: &mut RDKafka, oauthbearer_config: *const i8) {
         // set the token with rdkafka_sys::rd_kafka_oauthbearer_set_token
    }
}

let config = ClientConfig::new();
config.use_oauth_token_refresh_cb();
// other settings

let context = RefreshContext::new();
let producer = FutureProducer<RefreshContext> = config
    .create_with_context(context)
    .expect("Producer creation failed");

// etc.
```

Let me know what you think. The prototyping I've done for this has been on an older version of this crate, so testing is needed for these changes, particularly for `Consumer` types.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -11,6 +11,7 @@
 //! [`consumer`]: crate::consumer
 //! [`producer`]: crate::producer
 
+use std::borrow::Cow;
 use std::convert::TryFrom;
 use std::ffi::{CStr, CString};
 use std::mem::ManuallyDrop;
@@ -65,7 +66,7 @@ impl OAuthTokenData {
     }
 }
 /// Reporting mechanism for errors in generating OAuth tokens.
-pub struct OAuthTokenError(String);
+pub struct OAuthTokenError(pub String);
 
 /// Result type specifically used for generating OAuth tokens
 pub type OAuthResult = Result<OAuthTokenData, OAuthTokenError>;
@@ -515,7 +516,11 @@ pub(crate) unsafe extern "C" fn native_oauth_refresh_cb<C: ClientContext>(
 ) {
     // generate the token using generate_oauth_token
     let context = &mut *(opaque as *mut C);
-    let oauthbearer_config = CStr::from_ptr(oauthbearer_config).to_string_lossy();
+    let oauthbearer_config = match oauthbearer_config.is_null() {
+        true => Cow::from(""),
+        false => CStr::from_ptr(oauthbearer_config).to_string_lossy(),
+    };
+
     let token_info = match context.generate_oauth_token(oauthbearer_config.trim()) {
         Ok(token_info) => token_info,
         Err(OAuthTokenError(errmsg)) => {
@@ -542,7 +547,7 @@ pub(crate) unsafe extern "C" fn native_oauth_refresh_cb<C: ClientContext>(
         }
     };
 
-    let errstr = match CString::new(vec![0_u8; token_info.errstr_size]) {
+    let errstr = match CString::new(vec![u8::MAX; token_info.errstr_size]) {
         Ok(errstr) => errstr,
         Err(_) => {
             let errmsg = "Could not create error string";
--- a/src/client.rs
+++ b/src/client.rs
@@ -105,6 +105,15 @@ pub trait ClientContext: Send + Sync {
         error!("librdkafka: {}: {}", error, reason);
     }
 
+    /// Refreshes the OAuth token.
+    ///
+    /// The default implementation just logs an error message at the `error`
+    /// log level. This function is called if/when the token refresh callback
+    /// is called.
+    fn refresh_oauth_token(&self, _client: *mut RDKafka, _oauthbearer_config: *const i8) {
+        error!("Default token refresh only prints an error");
+    }
+
     // NOTE: when adding a new method, remember to add it to the
     // FutureProducerContext as well.
     // https://github.com/rust-lang/rfcs/pull/1406 will maybe help in the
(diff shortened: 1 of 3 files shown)


jsurany-bloomberg:
Should resolve issue https://github.com/fede1024/rust-rdkafka/issues/415

jsurany-bloomberg:
cc @benesch

benesch:
Sorry, I don't have time to look at this at the moment! I'll ping some of the other maintainers.

src/client.rs
```diff
@@ -105,6 +105,15 @@ pub trait ClientContext: Send + Sync {
         error!("librdkafka: {}: {}", error, reason);
     }
 
+    /// Refreshes the OAuth token.
+    ///
+    /// The default implementation just logs an error message at the `error`
+    /// log level. This function is called if/when the token refresh callback
+    /// is called.
+    fn refresh_oauth_token(&self, _client: *mut RDKafka, _oauthbearer_config: *const i8) {
```
duarten, line 113: Maybe `outhbearer_config` should have a more friendly type, like a `&[u8]`?

src/client.rs
```diff
@@ -447,6 +464,15 @@ pub(crate) unsafe extern "C" fn native_error_cb<C: ClientContext>(
     context.error(error, reason.trim());
 }
 
+pub(crate) unsafe extern "C" fn native_oauth_refresh_cb<C: ClientContext>(
+    client: *mut RDKafka,
+    oauthbearer_config: *const i8,
+    opaque: *mut c_void,
+) {
+    let context = &mut *(opaque as *mut C);
+    context.refresh_oauth_token(client, oauthbearer_config);
```
duarten, line 473: Should it return a boolean so we can then call `oauthbearer_set_token()` or 
`oauthbearer_set_token_failure()`?

src/client.rs
jsurany-bloomberg: I'd be willing to make it `&str` or `String` since the oauthbearer configuration is supposed to be `key=value` pairs.

For example, this will convert into `&str`
```rust
let oauthbearer_config = CStr::from_ptr(oauthbearer_config).to_string_lossy();
context.refresh_oauth_token(client, oauthbearer_config.trim());
```

src/client.rs
jsurany-bloomberg: Hoping that I'm understanding what you want here...

It'd be cool to put the unsafe calls to those functions in this block, but I'm not sure it makes sense.

For instance, I don't think we should assume what the value of parameters like `md_lifetime_ms` should be, and I don't think we should complicate `ClientContext` so that the user has a way to provide that value through context by default.

Also, I don't really want to assume anything about how the user handles errors in this process. By moving the call to `oauthbearer_set_token` into this block, we make it so that the user has a lot of flexibility in handling errors in generating the token (since that's code that they are going to write), but they have no flexibility in handling errors that get reported back via the `errstr` value - they would be bound to whatever we decide to do in this block.

I can write a pseudo prototype of a callback function to make things a bit clearer if needed.

src/client.rs
duarten: `&str` sound good!

src/client.rs
duarten: The callback could return the information needed for us to call `oauthbearer_set_token` here. One option would be to return `Result<SetTokenData, Error>`. The user can still have some custom error handling logic in the callback, but we'll take care of enforcing the [contract](https://docs.confluent.io/platform/current/clients/librdkafka/html/classRdKafka_1_1OAuthBearerTokenRefreshCb.html) that we must call `oauthbearer_set_token_failure` or `oauthbearer_set_token` in the callback.

Alternatively, if we really want the user to have all the flexibility in handling errors, we can return `Option<SetTokenData>`.

In any case, should this PR also include bindings for `oauthbearer_set_token_failure` or `oauthbearer_set_token`?

jsurany-bloomberg: Take a look at the most recent commit and let me know what you think. The only thing that's missing right now is support for SASL extensions, for which I would be happy to add support in my free time after this PR.

duarten: Looks great!

duarten (approved):

jsurany-bloomberg:
@duarten Tested these changes and discovered some issues. I've fixed them in the latest commit, along with a small QoL change to the `OAuthTokenError` struct.

src/client.rs
```diff
@@ -515,7 +516,11 @@ pub(crate) unsafe extern "C" fn native_oauth_refresh_cb<C: ClientContext>(
 ) {
     // generate the token using generate_oauth_token
     let context = &mut *(opaque as *mut C);
-    let oauthbearer_config = CStr::from_ptr(oauthbearer_config).to_string_lossy();
+    let oauthbearer_config = match oauthbearer_config.is_null() {
```
jsurany-bloomberg, line 519: `oauthbearer_config` may be null if there is no configuration. This check protects against the resulting seg fault.

src/client.rs
```diff
@@ -542,7 +547,7 @@ pub(crate) unsafe extern "C" fn native_oauth_refresh_cb<C: ClientContext>(
         }
     };
 
-    let errstr = match CString::new(vec![0_u8; token_info.errstr_size]) {
+    let errstr = match CString::new(vec![u8::MAX; token_info.errstr_size]) {
```
jsurany-bloomberg, line 550: [from documentation](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new) `CString::new` returns an error if the bytes contain a `0`, causing this part of the code to only error.

jsurany-bloomberg: `from_vec_unchecked` may be a better choice. I would think that we _want_ the vector to have only zeros.

PR #443, opened 2022-02-07
convert final log line to `crate::log`

guswynn:
I missed this one in https://github.com/fede1024/rust-rdkafka/pull/440

cc @benesch


benesch:
Thanks!

PR #454, opened 2022-04-03
Allow distinguishing between null and empty header values

benesch:
Kafka distinguishes between null header values and empty header values,
but rust-rdkafka was presenting null header values as empty header
values. Rework the API to preserve the distinction.


duarten (approved):

benesch:
Thanks for the review, @duarten!

PR #457, opened 2022-04-04
Rework OAuth callback to avoid memory unsafety and follow conventions

benesch:
@duarten @jsurany-bloomberg these changes are a result of me eyeballing the diff in #442. Any chance one of you is able to test? I don't have easy access to a Kafka broker configured for OAUTHBEARER authentication. Ideally we'd have a test for this in the repository, but that might be more trouble than it's worth to set up.

----

Tweak the support for the OAuth token refresh callback added in #442 to
avoid some memory unsafety and to be more in line with crate
conventions. In particular:

  * Don't pass `&str` to C code expecting a C string directly, as the
    Rust string slice is not guaranteed to be null terminated and so the
    C code might read off the end of a string. Instead always construct
    a CString.

  * Don't allow the user to specify an `errstr_size`. Crate convention
    is to use an `ErrBuf` with a fixed size of 512.

  * Rename `generate_oauth_token` to `refresh_oauth_token`, to hew
    closer to the librdkafka naming for the feature.

  * Move the `Config::use_token_refresh_cb` method to an associated
    constant on the `ClientContext`, so that the configuration of the
    token refresh callback is entirely contained within `ClientContext.`

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -11,8 +11,8 @@
 //! [`consumer`]: crate::consumer
 //! [`producer`]: crate::producer
 
-use std::borrow::Cow;
 use std::convert::TryFrom;
+use std::error::Error;
 use std::ffi::{CStr, CString};
 use std::mem::ManuallyDrop;
 use std::os::raw::{c_char, c_void};
@@ -31,44 +31,7 @@ use crate::groups::GroupList;
 use crate::log::{debug, error, info, trace, warn};
 use crate::metadata::Metadata;
 use crate::statistics::Statistics;
-use crate::util::{ErrBuf, KafkaDrop, NativePtr, Timeout};
-
-/// OAuthToken Data
-///
-/// When using token refresh, this data structure provides the fields used in
-/// the OAuth token refresh callback function.
-///
-/// NOTE: SASL extensions are not currently supported
-pub struct OAuthTokenData {
-    token: String,
-    lifetime_ms: i64,
-    principal_name: String,
-    errstr_size: usize,
-}
-
-impl OAuthTokenData {
-    /// Creates a new token data structure, with given value, lifetime, and
-    /// principal name. Error string size is set to 512 bytes.
-    pub fn new(token: String, lifetime_ms: i64, principal_name: String) -> Self {
-        Self {
-            token,
-            lifetime_ms,
-            principal_name,
-            errstr_size: 512,
-        }
-    }
-
-    /// Modifies the error string size to the specified value.
-    pub fn with_errorstring_size(&mut self, errstr_size: usize) -> &mut OAuthTokenData {
-        self.errstr_size = errstr_size;
-        self
-    }
-}
-/// Reporting mechanism for errors in generating OAuth tokens.
-pub struct OAuthTokenError(pub String);
-
-/// Result type specifically used for generating OAuth tokens
-pub type OAuthResult = Result<OAuthTokenData, OAuthTokenError>;
+use crate::util::{self, ErrBuf, KafkaDrop, NativePtr, Timeout};
 
 /// Client-level context.
 ///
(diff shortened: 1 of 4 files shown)


jsurany-bloomberg:
I can test the changes

src/client.rs
```diff
@@ -500,77 +483,73 @@ pub(crate) unsafe extern "C" fn native_error_cb<C: ClientContext>(
...
-        Err(_) => {
-            let errmsg = "Could not convert principal_name String to CString";
-            handle_refresh_error_msg(client, errmsg);
-            return;
+    let res: Result<_, Box<dyn Error>> = (|| {
+        let context = &mut *(opaque as *mut C);
+        let oauthbearer_config = match oauthbearer_config.is_null() {
+            true => None,
+            false => Some(util::cstr_to_owned(oauthbearer_config)),
+        };
```
duarten, line 512: nit: `(!oauthbearer_config.is_null()).then(|| util::cstr_to_owned(oauthbearer_config))`

duarten (approved):
Thanks for cleaning it up!

src/producer/future_producer.rs
```diff
@@ -153,9 +154,11 @@ impl<C: ClientContext + 'static> ClientContext for FutureProducerContext<C> {
         self.wrapped_context.error(error, reason);
     }
```
jsurany-bloomberg, line 156: Since `FutureProducerContext` wraps the inner context, you also need to wrap the constant

```rust
const ENABLE_REFRESH_OAUTH_TOKEN_CALLBACK: bool = C::ENABLE_REFRESH_OAUTH_TOKEN_CALLBACK;
```

Otherwise it falls back to the default, `false`

jsurany-bloomberg (changes requested):

src/producer/future_producer.rs
jsurany-bloomberg: the code that I'm using for my tests is using old `consumer` code, I don't really have the ability to test the consumer side of these changes in the current state of this repo, but from what I can tell it should be fine.

src/client.rs
```diff
@@ -500,77 +483,73 @@ pub(crate) unsafe extern "C" fn native_error_cb<C: ClientContext>(
...
-        error_msg.as_ptr() as *const c_char,
-    );
+/// A generated OAuth token and its associated metadata.
+///
+/// When using the `OAUTHBEARER` SASL authentication method, this type is
+/// returned from [`ClientContext::refresh_oauth_token`]. The token and
+/// principal name must not contain embedded null characters.
+///
+/// Specifying SASL extensions is not currently supported.
+pub struct OAuthToken {
```
jsurany-bloomberg, line 493: If/when SASL extensions support is added, is the plan to add another `pub` field to this struct? If so, then adding support for SASL extensions will break the public-facing API for this struct, which I'd rather not happen.

If I'm wrong, please let me know what your plan is. Otherwise, I think maybe we keep the fields `pub(crate)` at most, and use `OAuthToken::new(token: String, principal_name: String, lifetime_ms: i64)` as the public api.

src/client.rs
```diff
@@ -142,12 +115,22 @@ pub trait ClientContext: Send + Sync {
         error!("librdkafka: {}: {}", error, reason);
     }
 
-    /// Generates the OAuth token.
+    /// Refreshes an OAuth token from the provided configuration.
+    ///
+    /// Override with an appropriate implementation when using the `OAUTHBEARER`
+    /// SASL authentication mechanism. For this method to be called, you must
+    /// also set [`ClientContext::ENABLE_REFRESH_OAUTH_TOKEN_CALLBACK`] to true.
+    ///
```
jsurany-bloomberg, line 123: I don't think that this description does as good job of telling the user what they need to actually do for their implementation of this function. If I saw that the function name was `refresh_oauth_token`, I'd assume that I have to call `rd_kafka_oauthbearer_set_token` or `rd_kafka_oauthbearer_set_token_failure` in my implementation, which is not what we want.

I think we should be explicit that the expectation for this function is _only_ that the user create the token and return it (or some compatible error).

benesch:
My time has been in exceedingly short supply lately, sorry! I'll leave it up to someone else to address review feedback and merge, if desired.

vigneshs-12:
@benesch @jsurany-bloomberg Can we expect this changes to be available for use any soon ? / in next release?

src/producer/future_producer.rs
benesch: Good call, thank you!

src/client.rs
benesch: We're still pre-1.0, so we're free to make breaking changes at any time. I'm going to leave as is since it's quite a bit more convenient to have the named fields, and we can revisit if we ever get close to shipping a 1.0.

src/client.rs
benesch: Fair enough. I backed out the rename.

PR #459, opened 2022-04-06
Add detach_headers for OwnedMessage

scanterog:
Currently an `OwnedMessage` only allows to get a reference to the attached `OwnedHeaders`. This PR allows to detach the `OwnedHeaders` from the `OwnedMessage`.

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -517,6 +517,12 @@ impl OwnedMessage {
             headers,
         }
     }
+
+    /// Detach `OwnedHeaders` from OwnedMessage.
+    /// If you just need a reference, peek the headers() method.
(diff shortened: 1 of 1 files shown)


src/message.rs
```diff
@@ -517,6 +517,12 @@ impl OwnedMessage {
             headers,
         }
     }
+
+    /// Detach `OwnedHeaders` from OwnedMessage.
+    /// If you just need a reference, peek the headers() method.
```
duarten, line 522: I think we can omit this second sentence, and link to the `OwnedHeaders`:
```suggestion
    /// Detaches the [`OwnedHeaders`] from this `OwnedMessage`.
```

duarten (changes requested):

duarten (approved):

PR #461, opened 2022-04-08
Fix rdkafka-sys build for Bazel

guillaumebort:
When build with Bazel rules rust, there was 2 things not working here:
- The current directory when running this build script is the top level one
- `CARGO_MAKEFLAGS` env is not set

### DIFF
--- a/rdkafka-sys/build.rs
+++ b/rdkafka-sys/build.rs
@@ -72,6 +72,10 @@ fn main() {
             }
         }
     } else {
+        let rdkafkasys_root = Path::new("rdkafka-sys");
+        if rdkafkasys_root.exists() {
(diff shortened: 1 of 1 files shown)


rdkafka-sys/build.rs
```diff
@@ -72,6 +72,10 @@ fn main() {
             }
         }
     } else {
+        let rdkafkasys_root = Path::new("rdkafka-sys");
+        if rdkafkasys_root.exists() {
+            assert!(env::set_current_dir(&rdkafkasys_root).is_ok());
+        }
```
duarten, line 78: I don't think this should be here, the build tool should ensure the code is built from the right directory (as cargo does). This probably means a BUILD.bazel file is missing from this directory?

PR #462, opened 2022-04-11
Enable Atomic swap for statistics coming from librdkafka by enabling `Clone` on `Statistics`

vertexclique:
Adding `Clone` will enable sharing through atomic replacement with a callback.

How?
https://github.com/vertexclique/callysto/blob/master/src/kafka/contexts.rs#L31-L49

### DIFF
--- a/src/statistics.rs
+++ b/src/statistics.rs
@@ -15,7 +15,7 @@ use std::collections::HashMap;
 use serde::Deserialize;
 
 /// Overall statistics.
-#[derive(Deserialize, Debug, Default)]
+#[derive(Deserialize, Debug, Default, Clone)]
 pub struct Statistics {
     /// The name of the librdkafka handle.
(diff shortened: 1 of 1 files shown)


duarten:
Thanks for the PR, @vertexclique! I'm fine with adding `Clone` since it doesn't hurt and can be useful in other contexts, but looking at the linked code, isn't the appropriate fix to make `replace_with` accept a `FnMut` so you don't need to clone at all and can move from `cstat`?

vertexclique:
> isn't the appropriate fix to make replace_with accept a FnMut so you don't need to clone at all

It is, true, but as you can see if there is a lock it will be poisoned between reads of librdkafka and multiple threads that the framework uses to read. `RwLock` is also going to give stalls between readers and writers. So that's the more appropriate implementation for the library.

PR #467, opened 2022-05-02
WIP: Mock cluster support

GregBowyer:
This is an attempt to expose the mock cluster somewhat directly. This allows for control over the mock cluster, giving end users the ability to alter behaviors for a degree of fault injection.

I fully expect the APIs to change, and I am probably not quite maintaining the full style of the project.


GregBowyer:
Ping?

vojtechkral:
I was just looking into adding support for mocking and found your PR.

Would you be willing to add support for `rd_kafka_handle_mock_cluster()` too? That seems to be the easiest way to me to use mocking, without much code modification in an existing app, ie:

1. set `test.mock.num.brokers` to some number
2. obtain the mock cluster with `rd_kafka_handle_mock_cluster()`
3. get the bootstrap servers from the mock cluster
4. configure a counterpart producer/consumer to the app being tested

Probably this would be added on the `Client` type, ie. something like

```rust
fn Client::mock_cluster<'s>(&'s self) -> Option<MockCluster<'s>>;
```

which is annoying, because your existing `MockCluster` is an owning type... Maybe something similar to `PathBuf` and `Path<'_>` could be done where `PathBuf` dereferences to `&Path`?
Feel free to LMK what you think...

vojtechkral:
@GregBowyer here's a proposal: https://github.com/ticketmaster/rust-rdkafka/pull/1

GregBowyer:
@vojtechkral did add your pr into the mix, it seems sensible

vojtechkral:
@GregBowyer yes, thank you very much! I was meaning to ping the maintainer... but they're probably pretty busy...

benesch:
I'm afraid I'm not going to have time to look at this, but I'm 👍🏽 on adding it. Perhaps one of the other maintainers will be able to take a look.

davidblewett:
Closed in favor of https://github.com/fede1024/rust-rdkafka/pull/583 .

Issue #471, opened 2022-05-22
Possible unsoundness in IntoOpaque to do with pointer provenance and Box/Arc


zRedShift:
According to the `alloc` crate documentation, [`Arc::from_raw`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.from_raw) must be paired with a call to `Arc::into_raw`, otherwise pointer provenance may be violated (due to the `ArcInner` layout), and a slightly different issue with `Box`.

Regarding `Arc`, it could be solved by imitating `Arc::from_raw` and using `Arc::as_ptr`, as using the `Deref::deref`/`AsRef::as_ref` implementation might be [problematic](https://doc.rust-lang.org/1.61.0/src/alloc/sync.rs.html#849-851), but according to [this comment](https://doc.rust-lang.org/1.61.0/src/alloc/boxed.rs.html#1096-1100), there's some extra ceremony required for `Box` due to the uniqueness requirement.

I'm proposing to switch the `as_ptr` method to `into_ptr` and take `Self`. With that, `Box::into_raw` and `Arc::into_raw` could be used directly. This requires a small extra step to avoid a partial move, and incompatible with `!Sized`, but the fat pointer metadata can't be stored in `*mut c_void` anyway, and would require double boxing anyway (e.g. `Box<Box<dyn Trait>>`).

PR #472, opened 2022-05-22
Fix possible unsoundness in IntoOpaque to do with pointer provenance

zRedShift:
See #471

For `Arc`, an alternative that doesn't break the existing code would be to leave `as_ptr` as is and use `Arc::as_ptr(self)`, but the relevant `Box` methods take by `self`, so a more thorough analysis is required.

### DIFF
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -323,7 +323,7 @@ where
     // unstable.
     pub fn send<'a, K, P>(
         &self,
-        record: BaseRecord<'a, K, P, C::DeliveryOpaque>,
+        mut record: BaseRecord<'a, K, P, C::DeliveryOpaque>,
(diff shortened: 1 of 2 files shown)


src/producer/base_producer.rs
```diff
@@ -338,6 +338,7 @@ where
         let (payload_ptr, payload_len) = as_bytes(record.payload);
         let (key_ptr, key_len) = as_bytes(record.key);
         let topic_cstring = CString::new(record.topic.to_owned()).unwrap();
+        let ptr = record.delivery_opaque.into_ptr();
```
duarten, line 341: Can we give it a more descriptive name, like opaque_ptr?

duarten (commented):
LGTM, just a small nit.

PR #496, opened 2022-09-01
Add a new project using rust-rdkafka

awmatheson:
👋 A small change that adds in the README a new project that is using rust-rdkafka. Thanks for all the work on this.


benesch:
Thanks!

PR #486, opened 2022-06-27
Upgrade librdkafka to v1.9.0.

gyfis:
Resolves #485

PR #498, opened 2022-09-25
rdkafka-sys: make CARGO_MAKEFLAGS optional

mattbonnell:
Currently, the expectation that the `CARGO_MAKEFLAGS` env var be set causes the `rdkafka-sys/build.rs` script to fail when invoked in certain contexts -- notably, through https://github.com/bazelbuild/rules_rust rules. 

The `CARGO_MAKEFLAGS` env var is expected in order to populate the `MAKEFLAGS` env var. However, `MAKEFLAGS` is only required by GNU make. This commit refactors this section so that `CARGO_MAKEFLAGS` is only expected if GNU make is to be used.


benesch:
Thanks! Folding into #486.

PR #510, opened 2022-11-05
Support delete groups in the admin client

andrewinci:


benesch:
Code looks about right, but needs a test! Are you up for adding one?

andrewinci:
I've added a couple of tests, please let me know if there is any other case I should cover :)

benesch:
Pushed up a few fixes. Will merge as #511, since I don't have access to push at this PR.

Issue #509, opened 2022-10-31
Seeing an issue where a `StreamConsumer` hangs indefinitely in `BaseConsumer<C>::drop()`


penick:
I'm able to fairly regularly able to produce the issue using this sample program: https://github.com/penick/kafka_hang/blob/main/src/main.rs. This hang also manifests itself in the [Kafka Source](https://github.com/vectordotdev/vector/blob/v0.24.2/src/sources/kafka.rs) for Vector.

Versions: v0.28.0 and v0.29.0

Also, I noticed this comment in the [code](https://github.com/fede1024/rust-rdkafka/blob/v0.29.0/src/consumer/base_consumer.rs#L549):

```rust
trace!("Destroying consumer: {:?}", self.client.native_ptr()); // TODO: fix me (multiple executions ?)
```

I was wondering what that meant as it might have some relevance here, but I'm not sure.  After looking in the `librdkafka` code I think it may be possible to hang `rd_kafka_consumer_close()` if called in rapid succession. 

This is the stack trace when it hangs:

```
Thread 9 (Thread 0x7fe4cb0a4640 (LWP 50640) "tokio-runtime-w"):
#0  __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7fe4bc034a90) at ./nptl/futex-internal.c:57
#1  __futex_abstimed_wait_common (cancel=true, private=0, abstime=0x0, clockid=0, expected=0, futex_word=0x7fe4bc034a90) at ./nptl/futex-internal.c:87
#2  __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0x7fe4bc034a90, expected=expected@entry=0, clockid=clockid@entry=0, abstime=abstime@entry=0x0, private=private@entry=0) at ./nptl/futex-internal.c:139
#3  0x00007fe4cbf59ac1 in __pthread_cond_wait_common (abstime=0x0, clockid=0, mutex=0x7fe4bc034a40, cond=0x7fe4bc034a68) at ./nptl/pthread_cond_wait.c:503
#4  ___pthread_cond_wait (cond=0x7fe4bc034a68, mutex=0x7fe4bc034a40) at ./nptl/pthread_cond_wait.c:627
#5  0x00007fe4cbf636fd in __cnd_wait (cond=<optimized out>, mutex=<optimized out>) at ../sysdeps/pthread/cnd_wait.c:25
[...]

PR #517, opened 2022-11-23
fix: dropping a consumer can block indefinitely

dhable:
The call to `rd_kafka_consumer_close()` in the `Drop` trait of `BaseConsumer` can block in the rebalance callback indefinitely. It's stuck in the call `rd_kafka_rebalance_protocol()` waiting for a queue operation to be completed. We often see this happen when an external process deletes a topic that was currently being consumed in our application, leaving thread trying to drop the `BaseConsumer` in a hung state.

Rust drops fields in a struct in the order they are declared in the source. When the `StreamConsumer` is dropped, it will first call `base.drop()`, which eventually executes `rd_kafka_consumer_close()`, and then calls `queue.drop()`, which eventually executes `rd_kafka_queue_destroy()`. This ordering is incorrect according to the documentation for [rd_kafka_queue_get_consumer()](https://docs.confluent.io/3.2.1/clients/librdkafka/rdkafka_8h.html#acacdb55ae7cb6abfbde89621e512b078). Moving the queue `field` above `base` causes rust to then call the underlying librdkafka functions in the correct order when the `StreamConsumer` is dropped.

Shout out to @penick for the collaboration on troubleshooting and testing the fix.

### DIFF
--- a/src/consumer/stream_consumer.rs
+++ b/src/consumer/stream_consumer.rs
@@ -165,9 +165,9 @@ pub struct StreamConsumer<C = DefaultConsumerContext, R = DefaultRuntime>
 where
     C: ConsumerContext,
 {
+    queue: NativeQueue, // queue must be dropped before the base to avoid deadlock
     base: BaseConsumer<C>,
     wakers: Arc<WakerSlab>,
-    queue: NativeQueue,
     _shutdown_trigger: oneshot::Sender<()>,
     _runtime: PhantomData<R>,
 }


penick:
fixes: https://github.com/fede1024/rust-rdkafka/issues/509

duarten (approved):
Nice find!

benesch:
Thanks very much for tracking this down!

yankm:
That was a great find indeed, thank you for the fix! @benesch could you please make a release with it included? We are currently hitting this problem in production.

PR #520, opened 2022-12-04
Allow purging (Purge API) / don't leak on dropping producer

### LINKED ISSUE #518: Allow purging (purge API) / don't leak on dropping producer
I'm looking for a way to handle messages that have failed to send upon closing.

Currently, dropping the producer when kafka server is not started gives the following message:
```
[WARN] librdkafka: TERMINATE [thrd:app]: Producer terminating with 2 messages (13 bytes) still in queue or transit: use flush() to wait for outstanding message delivery
```

Flushing beforehand (`ThreadedProducer::flush`) will obviously always timeout because server is not running, and doesn't change the previous message.

Delivery callbacks from `ProducerContext` do not get called for these messages, leading to memory leaks (as Box/Arc/... are not cast back).

librdkafka provides the *purge API* to solve this issue (https://github.com/edenhill/librdkafka/issues/990, https://github.com/edenhill/librdkafka/commit/107af2a5c9c564b71b06266227e76f1f8fad942c#diff-84ac6df9f4f6c13f1f4eecf3b9c0e2cee4d7c4c005feea512ff3f82c000fb84eR2211).

It would be useful to have access to this API for scenarios such as mine or those described at https://github.com/edenhill/librdkafka/issues/990.

As the Rust API also does some nice `IntoOpaque` casting where it may leak if callbacks are not called, I think it would also be best to actually force an instant purge/flush when dropping the `BaseProducer`, both ensuring that the `DeliveryOpaque`s don't get somewhat-silently leaked, and making it more natural for users to handle failure to deliver messages.
(The reason this wasn't done in librdkafka [is that users may already be handling this separately](https://github.com/edenhill/librdkafka/issues/990#issuecomment-300587671), but that could just be released in 0.30.0 which would allow breaking changes - and this is unlikely to break much because such handling would be severely un-idiomatic as people would have to keep track of the set of raw pointers that were sent, removing them from the list as they get a callback, and cast them back using unsafe code to drop them in case they didn't receive the callback, instead of just opening the issue I'm opening right now)

Thanks!

Ten0:
Resolves #518

**What it does:**
- Purges producer on drop
- Updates bindgen script to more recent bindgen version
- Sets the parameter that makes macro constants signed (because that's how they are generally used in librdkafka)

Todo:
- [x] Check that it works
- [x] Expose the full interface


Ten0:
Could use a ci run as well 🙂

Ten0:
This is ready.

I [went ahead](https://github.com/fede1024/rust-rdkafka/issues/518#issuecomment-1336480692) and pulled the trigger on removing the `Clone` bound on the `BaseProducer` in ea0e6858bfc6d173d337fc3d914dabfbf75094c4 because it appears to really be unnecessary and overcomplexifies stuff for no gain.

It seems similarly unnecessary on the `FutureProducer` at the moment, but because there's runtime awareness with spawn capabilities I could actually imagine that it may be useful to be able to spawn tasks that refer to it some day, and it doesn't really complexify anything there (it doesn't need a particular `Drop` impl) so it seems reasonable to leave internal reference counting there for now.

Ten0:
@benesch 
> If you've got the time, please submit a PR! The maintainers don't have any spare bandwidth these days, I'm afraid.

Since you suggested that I dedicate time to implementing this... when should I hope for this to be reviewed?

rdkafka-sys/update-bindings.sh
```diff
@@ -9,13 +9,14 @@ bindgen \
...
-    --blacklist-function "rd_kafka_conf_set_open_cb" \
-    --raw-line "use libc::{FILE, sockaddr, size_t, ssize_t, c_int, c_void, c_char};" \
+    --allowlist-function "rd_kafka.*" \
+    --allowlist-type "rd_kafka.*" \
+    --allowlist-var "rd_kafka.*|RD_KAFKA_.*" \
+    --no-recursive-allowlist \
+    --blocklist-function "rd_kafka_conf_set_open_cb" \
+    --raw-line "use libc::{FILE, sockaddr, c_int, c_void, c_char};" \
     --raw-line "use num_enum::TryFromPrimitive;" \
+    --default-macro-constant-type "signed" \
```
benesch, line 19: Was this important for this PR? Or just avoiding an `as i32` somewhere?

src/producer/mod.rs
```diff
@@ -368,3 +387,30 @@ where
...
+        /// broker, making it impossible for the application to know if these
+        /// messages were successfully delivered or not.
+        /// Retrying these messages may lead to duplicates.
+        ///
+        /// This does not purge messages in internal queues.
+        const INFLIGHT = rdkafka_sys::RD_KAFKA_PURGE_F_INFLIGHT;
+        /// Don't wait for background thread queue purging to finish.
+        const NON_BLOCKING = rdkafka_sys::RD_KAFKA_PURGE_F_NON_BLOCKING;
+    }
+}
```
benesch, line 416: I feel like it would be Rust-ier to use a `PurgeConfig` struct here with `queue(bool)`, `in_flight(bool)`, and `non_blocking(bool)` methods. No particular need to save the two bytes. Also avoids the dependency on the `bitflags` crate, which idk someone will probably care about. (We get a lot of PRs to minimize crate dependencies.)

benesch (commented):
I'm afraid I only have an hour or two every few months to spend on rust-rdkafka these days.

This looks pretty solid. Thanks again for sending it up. Just one comment about the use of `bitflags`.

rdkafka-sys/update-bindings.sh
Ten0: It was avoiding an `as i32` in my code as well as in  existing code. Overall this seems to be how macro constants are generally used in librdkafka, so that seems to be the right way to represent these.

src/producer/mod.rs
Ten0: Done in 37e318097d8238eff430af5c301c5972bd35c935

davidblewett:
@Ten0 we're preparing a release of `v0.30.0`. Can you update this PR to fix the merge conflicts so we can include it?

Ten0:
done

scanterog (approved):

PR #528, opened 2022-12-22
scram-vendored feature

sproberts92:
sasl2-sys now has a separate feature for SCRAM without krb5. We'd like to expose this as a feature so we can use it downstream in https://github.com/vectordotdev/vector.

See the [issue in that project for more context](https://github.com/vectordotdev/vector/issues/14874). 

Credit to @ilmarii in that issue for figuring out what is actually needed here.


Neustradamus:
Thanks a lot @sproberts92!

benesch:
I don't think this works quite the way you think! SASL SCRAM authentication is enabled by the `ssl` or `ssl-vendored` features. The reason is that librdkafka actually has a built-in implementation of SASL SCRAM authentication, but it requires the base64 encoding routines from OpenSSL. So you need to enable SSL in order to get access to that built-in implementation.

librdkafka only uses libsasl2 for Kerberos authentication. That's why it's gated here under `gssapi`, rather than `libsasl2`.

tl;dr you probably just need to enable the `ssl-vendored` feature on Windows in Vector.

sproberts92:
@benesch thank you for your insight! I did not realise that librdkafka contained its own implementation for SCRAM, but indeed I now see it in there when I look through [confluentinc/librdkafka](https://github.com/confluentinc/librdkafka).

I'll close this for now and discuss with the others what we need back on our Vector issue.

PR #548, opened 2023-02-16
Fix truncation of `rdkafka_sys::rd_kafka_version` in `rdkafka::get_rdkafka_version`.

davidblewett:
I discovered this when I tried to add  runtime library version checking in our application. `rdkafka::get_rdkafka_version` silently truncates the integer representation so that it's impossible to properly handle how upstream represents the version ( https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#aa2e242fb8620a32b650a40575bc7f98e ):

```
Interpreted as hex MM.mm.rr.xx:

MM = Major
mm = minor
rr = revision
xx = pre-release id (0xff is the final release)
E.g.: 0x000801ff = 0.8.1
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,17 @@ See also the [rdkafka-sys changelog](rdkafka-sys/changelog.md).
 
 ## Unreleased
 
+* **Breaking change.** `util::get_rdkafka_version` now returns `(i32, String)`.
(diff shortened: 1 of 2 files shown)


benesch (approved):
Whoops! `as` is full of these sorts of surprises. Thanks for fixing.

PR #555, opened 2023-03-09
Add `Timestamp` to the `OwnedDeliveryResult` tuple.

gorzell:
**Note:** This is an API breaking change.

Currently the `Timestamp` value is being dropped as part of the conversion from `BorrowedMessage` to `OwnedDeliveryResult`, however this value is updated once the message has been published and callers should have access to the updated value. As far as I can tell, this is the **only** other value that is updated as part of publishing.

An alternative to expanding the tuple would be to replace it with a struct. This is a bit more disruptive now, but would making adding values in the future a backwards compatible change. So it depends how likely it is that there will be further new values.

This could also be done in a way that isn't a breaking change, but would involve duplicating functions and copying some types, i.e. `DeliveryFuture`. It doesn't seem like that level of complexity is justified given the pre 1.0 status.

### DIFF
--- a/src/producer/future_producer.rs
+++ b/src/producer/future_producer.rs
@@ -134,7 +134,7 @@ pub struct FutureProducerContext<C: ClientContext + 'static> {
 /// partition and offset of the message. If the message failed to be delivered
 /// an error will be returned, together with an owned copy of the original
 /// message.
-pub type OwnedDeliveryResult = Result<(i32, i64), (KafkaError, OwnedMessage)>;
+pub type OwnedDeliveryResult = Result<(i32, i64, Timestamp), (KafkaError, OwnedMessage)>;
 
 // Delegates all the methods calls to the wrapped context.
 impl<C: ClientContext + 'static> ClientContext for FutureProducerContext<C> {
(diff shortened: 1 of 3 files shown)


gorzell:
Any chance someone could take a look at this? I would like get our production systems off my fork if at all possible.

CleanCut:
@davidblewett When you get a moment, could you review this PR? We'd love to stop maintaining a fork for this small change. ❤️

CleanCut:
Thanks, @davidblewett! Good to know. We'll get back to you soon with that change.

CleanCut:
> We'll get back to you soon with that change.

Just to set expectations, we'll most likely deliver this change next month after the holidays. We want to make sure to give it a thorough review and testing on our end before submitting it upstream. Happy Holidays! 😄

gorzell:
> I would prefer a dedicated struct, for the reason described in the comment. If we're going to break the API, I'd prefer to do it once and be able to adjust it in the future without breaking users.

I have switched this out. Let me know if there is anything else you would like me to change.

CleanCut:
@davidblewett I hope you had some fantastic winter holidays! How is this PR looking to you?

davidblewett (approved):
+1

davidblewett (changes requested):
Last request I promise: please add a changelog entry documenting the API break for the `OwnedDeliveryResult` type.

gorzell:
@davidblewett changelog updated.

PR #567, opened 2023-04-11
Expose unassign

abolibibelot:
The librdkafka `assign` function can be used to clear the list of current assignments by passing null, but this wasn't exposed.

We added an explicit unassign method (similar to what [the cpp](https://github.com/confluentinc/librdkafka/blob/master/src-cpp/KafkaConsumerImpl.cpp#L181) and [confluent kafka go](https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/consumer.go#L151) clients do) which precisely does that.

Also exposed `incremental_assign` and `incremental_unassign` which were left off.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,12 @@ See also the [rdkafka-sys changelog](rdkafka-sys/changelog.md).
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


davidblewett (approved):
LGTM!

PR #570, opened 2023-04-25
update CI workflow to avoid deprecation warnings and build errors

ClementTsang:
I was taking a look at the repo and noticed that there are a lot of warnings regarding deprecated Node and Action commands regarding the CI workflow.

Part of resolving this is addressing `actions-rs/toolchain`, which is no longer maintained from my understanding. [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain) is actively maintained and works well from my experience, so I think switching over should be fine and resolves the issues. The rest is fixed by just updating `checkout` to `v3`.

This also updates the `rust_version` to 1.54.0, since [this change](https://github.com/fede1024/rust-rdkafka/commit/611aeea5012787ad0f1a3ff74af32b64d41142f8#diff-8731dd08efd465ecbfdd192013294eeb858fd4e18dbd44d7528dc7f763fbf108R440) means that 1.53.0 will no longer build. Not sure if this should be reflected in the MSRV.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
     branches: [master]
 
 env:
-  rust_version: 1.53.0
+  rust_version: 1.54.0
 
 jobs:
   lint:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,81 +13,78 @@ jobs:
   lint:
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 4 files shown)


ClementTsang:
[Sample run](https://github.com/ClementTsang/rust-rdkafka/actions/runs/4793271845) from my fork.

davidblewett (approved):
LGTM. I was working on this separately. Will merge after CI completes successfully.

PR #577, opened 2023-05-16
Support bazel builds

mborst:
Currently, bzl executes the build script in the wrong directory and fails. This checks for this case and sets the right directory if necessary.


davidblewett (approved):
+1

PR #586, opened 2023-06-13
Snappy is not enabled by default in cmake build, enable it

Magister:
For some reason librdkafka's cmake build does not enable snappy compression support.
This enables snappy unconditionally (because anyway it is not an external dependency)

### DIFF
--- a/rdkafka-sys/build.rs
+++ b/rdkafka-sys/build.rs
@@ -273,6 +273,8 @@ fn build_librdkafka() {
         config.define("ENABLE_LZ4_EXT", "0");
     }
 
+    config.define("WITH_SNAPPY", "1");
+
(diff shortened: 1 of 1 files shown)


chklauser:
Lack of snappy compression is an issue for using rust-rdkafka in my organisation. While we deploy to linux, where we could use the mklove-based librdkafka-build, our developers mostly run Windows, where only the CMake-based librdkafka-build is available.

I'm not familiar at all with the mklove build system, but from what I can tell, the mklove system also enables snappy by default (hard-coded).

What would it take for this to get merged? Probably an entry in the [changelog](../blob/master/changelog.md)? I'm happy to help (obviously can't push to this PR directly)

PR #587, opened 2023-06-13
Fix panic on getting config value from NativeConfig

Magister:
Kafka can return string with multiple `\0` chars (seen at least on Windows x64), and `CStr::from_bytes_with_nul` panics in that case. `String::from_utf8_lossy()` handles that ok

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -150,10 +150,7 @@ impl NativeClientConfig {
         }
 
         // Convert the C string to a Rust string.
-        Ok(CStr::from_bytes_with_nul(&buf)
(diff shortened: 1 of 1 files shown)


jjh42:
This change seems to result in getting an extra null character and therefore panicing (on linux) when trying to parse the result into a number.

The change in the linked PR fixes it (although that PR also contains a lot of other changes)

davidblewett:
@jjh42 yes; it might be simpler to revert this. I ran out of time ironing out the issues upgrading the MSRV to be able to handle this in a better fashion.

PR #592, opened 2023-06-23
Introduce partitioner interface

mlowicki:
Add interface to specify custom partitioners by extending `ProducerContext` trait with capability to return optional custom partitioner.

### DIFF
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -272,7 +272,7 @@ where
             unsafe {
                 rdsys::rd_kafka_topic_conf_set_opaque(
                     default_topic_config,
-                    Arc::as_ptr(&context) as *const Part as *mut c_void,
+                    Arc::as_ptr(&context) as *mut c_void,
                 )
             };
             unsafe {
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -234,12 +234,10 @@ unsafe extern "C" fn partitioner_cb<Part: Partitioner, C: ProducerContext<Part>>
 
     let producer_context = &mut *(rkt_opaque as *mut C);
 
-    match producer_context.get_custom_partitioner() {
-        None => panic!("custom partitioner is not set"),
-        Some(partitioner) => {
-            partitioner.partition(topic_name, key, partition_cnt, is_partition_available)
-        }
-    }
+    producer_context
+        .get_custom_partitioner()
+        .expect("custom partitioner is not set")
+        .partition(topic_name, key, partition_cnt, is_partition_available)
 }
 
 impl FromClientConfig for BaseProducer<DefaultProducerContext> {
@@ -266,7 +264,7 @@ where
         let native_config = config.create_native_config()?;
         let context = Arc::new(context);
 
-        if let Some(_) = context.get_custom_partitioner() {
+        if context.get_custom_partitioner().is_some() {
             let default_topic_config =
                 unsafe { rdsys::rd_kafka_conf_get_default_topic_conf(native_config.ptr()) };
             unsafe {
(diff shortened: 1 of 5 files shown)


scanterog:
Thanks for contributing. I've reviewed this and I personally think it would be better to use an associated type instead of generic type for the `Partitioner` in the `ProducerContext`.

I've spent some time and [created this patch](https://gist.github.com/scanterog/d6881ea908acb6c892c2cea1025629b6) so you can observe the difference. By doing that we avoid introducing the Partitioner trait pretty much everywhere.

I'm going on vacation next week and I'll let you sync with @duarten and @davidblewett on what you all think is better for this.

mlowicki:
> Thanks for contributing. I've reviewed this and I personally think it would be better to use an associated type instead of generic type for the `Partitioner` in the `ProducerContext`.
> 
> I've spent some time and [created this patch](https://gist.github.com/scanterog/d6881ea908acb6c892c2cea1025629b6) so you can observe the difference. By doing that we avoid introducing the Partitioner trait pretty much everywhere. The patch might require further testing. It was mostly to illustrate the idea. Ideally we also want a separate `CollectingContext` context for testing the custom partitioner on the low level producer tests.
> 
> I'm going on vacation next week and I'll let you sync with @duarten and @davidblewett on what you all think is better for this.

Yeah, so using associated type was my first approach (I've working version somewhere). Later on decided to use generic type because of few reasons:
* Associated type defaults are unstable (https://github.com/rust-lang/rust/issues/29661) so for every implementation you would need to add something like `type CustomPartitioner = NoCustomPartitioner;`. With generic types we can use default value like `pub struct ThreadedProducer<C, Part: Partitioner = NoCustomPartitioner>` so things will remain backward compatible. If we would go with associated types then it'll break for everyone - it won't be hard to fix it but still.
* It makes more sense to support more than one partitioner per ProducerContext. Let's take a very simple example like https://github.com/fede1024/rust-rdkafka/blob/master/tests/test_low_producers.rs#L71 (`CollectingContext`). Using Generic types allows for better code re-use.

For sure using generic types requires more changes in the code base + signatures are becoming more complex. I'll start a thread on Slack to discuss on the direction here.

src/producer/mod.rs
```diff
@@ -199,6 +199,56 @@ pub trait ProducerContext: ClientContext {
     /// failed to). The `DeliveryOpaque` will be the one provided by the user
     /// when calling send.
     fn delivery(&self, delivery_result: &DeliveryResult<'_>, delivery_opaque: Self::DeliveryOpaque);
+
+    /// This method is called when creating producer in order to register custom partitioner.
+    /// Box is used to make sure data is on the heap as partitioner address will fly across FFI boundary.
+    #[allow(clippy::borrowed-box)]
+    fn get_custom_partitioner(&self) -> Option<&Box<Part>> {
```
duarten, line 206: I think this should `Option<&Part>`.

src/producer/base_producer.rs
```diff
@@ -206,23 +213,80 @@ impl<'a, K: ToBytes + ?Sized, P: ToBytes + ?Sized> BaseRecord<'a, K, P, ()> {
...
+    let key = if keydata.is_null() {
+        None
+    } else {
+        Some(slice::from_raw_parts(keydata as *const u8, keylen))
+    };
+
+    let producer_context = &mut *(rkt_opaque as *mut C);
+
+    match producer_context.get_custom_partitioner() {
+        None => panic!("custom partitioner is not set"),
```
davidblewett, line 238: Hmm; is it possible to validate that the custom partitioner is set before the callback is called? Or is this a "will never happen" to satisfy the type system?

I'm not a fan of panicing in a library. What is the effect if the callback was set but no actual partitioner was configured? Can we set an error state somewhere that's bubbled out to the caller to handle? If it's basically a config error I think logging an error is better than panicing.

mlowicki: Nothing stops author of producer context to e.g. return None from get_custom_partitioner() during 2nd and next call so in theory it can happen but it’s rather contrived case.

My first version there was to return PARTITION_UA (-1) but it leads to “message timeout” errors when message.timeout.ms elapses so for me that was more confusing than simply panicking right away.

akhil-karanth (approved):

PR #596, opened 2023-07-05
Change signature for seek_partitions

scanterog:
Following librdkafka, individual partition errors should be reported in the per-partition`error` field of `TopicPartitionList` elements.

Example where errors are returned for each partition rather than the method call (which succeded)
`TopicPartitionList {test-seek/0: offset=Offset(3) metadata="", error=Err(KafkaError (Offset fetch error: InProgress (Local: Operation in progress))); test-seek/1: offset=Offset(0) metadata="", error=Err(KafkaError (Offset fetch error: InProgress (Local: Operation in progress))); test-seek/2: offset=Offset(10) metadata="", error=Err(KafkaError (Offset fetch error: InProgress (Local: Operation in progress)))}`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,10 @@ See also the [rdkafka-sys changelog](rdkafka-sys/changelog.md).
 
 ## Unreleased
 
+* **Breaking change.** Change signature for `seek_partitions`. Following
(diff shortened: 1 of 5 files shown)


davidblewett (approved):
LGTM

PR #602, opened 2023-08-03
Allow users to override make when using cmake

joelwachsler:
For some reason the wrong make program is used when building using bazel. Making it possible to change the make program allows me to override it which fixes the problem.


davidblewett:
LGTM.

PR #616, opened 2023-10-20
Derives `serde::Serialize` on `Statistics`

icdevin:
I realize that there is a `stats_raw` trait method, but it would be better to not have to re-type out the statistics fields myself. This came about because I would like to act on many of the individual fields locally via some aggregations, but then output them back out (along with other fields in a larger status struct) as a JSON-encoded string, e.g.:

```rs
#[derive(Serialize)]
pub struct Status {
  pub sample_field: u64,
  pub sample_field_2: boolean,
  pub statistics: Statistics,
}
```

### DIFF
--- a/src/statistics.rs
+++ b/src/statistics.rs
@@ -12,7 +12,7 @@
 
 use std::collections::HashMap;
 
-use serde::{Serialize, Deserialize};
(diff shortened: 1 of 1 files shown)


icdevin:
I'm not sure that the failing test is related to my changes.

PR #617, opened 2023-10-26
Use rdkafka event API instead of the callback API

scanterog:
This PR changes how rust-rdkafka interacts with librdkafka. Instead of using the callback API, it moves to the event API. This PR does not change any of the public rdkafka API, but it give us the chance to introduce some optimizations. Some of them might break the current API though and hence this PR does not go into that route yet. For example, consider https://github.com/fede1024/rust-rdkafka/issues/273.

Some other benefits:
* it allows to extend the lifetime of BorrowMessages. For example, consider the `FutureProducer` where we need to [detach](https://github.com/fede1024/rust-rdkafka/blob/93e6cd4f36d04e01fe5e025699bb78be938c1f1a/src/producer/future_producer.rs#L187) the borrowMessage on errors.
* it allows handling the rebalance events outside of the scope of the registered callback. This is specially useful for services using rust-rdkafka to build FFI libraries on top of it.

Performance wise, I've run kafka-benchmark I've got a pretty similar outcome for both the callback vs event based approach. I didn't find much differences here. I've also run long running services on our staging infra and we didn't find noticeable changes on cpu/rss on the producer/consumer side.

There's one main difference here though. Operation errors not related to consumption (like transient broker transport failures) are passed back to the user with the Event API. This is not the case for the Callback API. Details can be found https://github.com/confluentinc/librdkafka/issues/4493.  These errors should generally be considered informational as the underlying client will automatically try to recover from any errors encountered, the application does not need to take action on them. However, fatal errors are also propagated via the Event API. These indicates that the particular instance of the client (producer/consumer) becomes non-functional.

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -113,7 +113,15 @@ where
     ///
     /// The returned message lives in the memory of the consumer and cannot outlive it.
     pub fn poll<T: Into<Timeout>>(&self, timeout: T) -> Option<KafkaResult<BorrowedMessage<'_>>> {
-        let event = self.client().poll_event(self.get_queue(), timeout.into())?;
+        self.poll_queue(self.get_queue(), timeout)
+    }
+
+    pub(crate) fn poll_queue<T: Into<Timeout>>(
+        &self,
+        queue: &NativeQueue,
+        timeout: T,
+    ) -> Option<KafkaResult<BorrowedMessage<'_>>> {
+        let event = self.client().poll_event(queue, timeout.into())?;
         let evtype = unsafe { rdsys::rd_kafka_event_type(event.ptr()) };
         match evtype {
             rdsys::RD_KAFKA_EVENT_FETCH => self.handle_fetch_event(event),
@@ -743,26 +751,7 @@ where
     /// associated consumer regularly, even if no messages are expected, to
     /// serve events.
     pub fn poll<T: Into<Timeout>>(&self, timeout: T) -> Option<KafkaResult<BorrowedMessage<'_>>> {
-        let event = self
-            .consumer
-            .client()
-            .poll_event(&self.queue, timeout.into())?;
-        let evtype = unsafe { rdsys::rd_kafka_event_type(event.ptr()) };
-        match evtype {
-            rdsys::RD_KAFKA_EVENT_FETCH => unsafe {
-                NativePtr::from_ptr(rdsys::rd_kafka_event_message_next(event.ptr()) as *mut _)
-                    .map(|ptr| BorrowedMessage::from_client(ptr, Arc::new(event), self))
-            },
-            _ => {
-                let buf = unsafe {
-                    let evname = rdsys::rd_kafka_event_name(event.ptr());
-                    CStr::from_ptr(evname).to_bytes()
-                };
-                let evname = String::from_utf8(buf.to_vec()).unwrap();
-                warn!("Ignored event '{}' on consumer poll", evname);
-                None
-            }
-        }
+        self.consumer.poll_queue(&self.queue, timeout)
     }
 
     /// Sets a callback that will be invoked whenever the queue becomes
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -113,9 +113,7 @@ where
     ///
     /// The returned message lives in the memory of the consumer and cannot outlive it.
     pub fn poll<T: Into<Timeout>>(&self, timeout: T) -> Option<KafkaResult<BorrowedMessage<'_>>> {
-        let event = self
-            .client()
-            .poll_event(self.get_queue(), timeout.into())?;
+        let event = self.client().poll_event(self.get_queue(), timeout.into())?;
         let evtype = unsafe { rdsys::rd_kafka_event_type(event.ptr()) };
         match evtype {
             rdsys::RD_KAFKA_EVENT_FETCH => self.handle_fetch_event(event),
@@ -145,7 +143,7 @@ where
     ) -> Option<KafkaResult<BorrowedMessage<'_>>> {
         unsafe {
             NativePtr::from_ptr(rdsys::rd_kafka_event_message_next(event.ptr()) as *mut _)
-                .map(|ptr| BorrowedMessage::from_client(ptr, event, self.client()))
+                .map(|ptr| BorrowedMessage::from_client(ptr, Arc::new(event), self.client()))
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/consumer/base_consumer.rs
```diff
@@ -607,8 +714,19 @@ where
...
+        trace!("Destroying consumer: {:?}", self.client.native_ptr());
+        if self.group_id.is_some() {
+            let err = unsafe {
+                rdsys::rd_kafka_consumer_close_queue(self.client.native_ptr(), self.queue.ptr())
+            };
+            if !err.is_null() {
+                error!("Failed to close the consumer queue on drop");
+            }
+
+            while unsafe { rdsys::rd_kafka_consumer_closed(self.client.native_ptr()) } != 1 {
```
davidblewett, line 726: Should this use `BaseConsumer::closed()`?

mensfeld:
Hey, could you explain to a fellow maintainer (rdkafka-ruby here) what benefits it yields for the consumer? I clearly see benefits for admin and producer (where the delivery reports materialization is easier to handle) but I wonder about the consumer. I would appreciate some explanation as it has been a while since I started thinking about this but I have problems wrapping my head around the consumer part. Is it not possible to yield from rust in the context of rebalance callback? How will it operate on rebalanes where the revocation happens first and prior to it an action like offset storage could take place? Thanks :pray:

scanterog:
> Is it not possible to yield from rust in the context of rebalance callback? How will it operate on rebalanes where the revocation happens first and prior to it an action like offset storage could take place?

Hello @mensfeld! The rebalance handler does not really change here for rust-rdkafka. Is there anything in particular that gives that impression? Can you elaborate a little bit more about this?

src/consumer/base_consumer.rs
```diff
@@ -96,66 +63,181 @@ where
...
+                            let evname = rdsys::rd_kafka_event_name(event.ptr());
+                            CStr::from_ptr(evname).to_bytes()
+                        };
+                        let evname = String::from_utf8(buf.to_vec()).unwrap();
+                        warn!("Ignored event '{}' on consumer poll", evname);
+                    }
+                }
+            }
+
+            if op_timeout >= timeout {
```
duarten, line 161: We set `op_timeout` to `std::cmp::min(timeout, min_poll_interval)`, so it's never going to be larger than `timeout`. Shouldn't we do something like:

```rust
let now = Instant::now();
let op_timeout = std::cmp::min(timeout, min_poll_interval);
// ...
let timeout = timeout.saturating_sub(now.elapsed());
if timeout == 0 {
    return None;
}
```

src/consumer/base_consumer.rs
```diff
@@ -607,8 +714,16 @@ where
...
-        trace!("Destroying consumer: {:?}", self.client.native_ptr()); // TODO: fix me (multiple executions ?)
-        unsafe { rdsys::rd_kafka_consumer_close(self.client.native_ptr()) };
+        trace!("Destroying consumer: {:?}", self.client.native_ptr());
+        if self.group_id.is_some() {
+            if let Err(err) = self.close_queue() {
+                error!("Failed to close consumer queue on drop: {}", err);
+            } else {
+                while !self.closed() {
+                    self.poll(Duration::from_millis(100));
+                }
```
duarten, line 724: Is this safe? Could we drop messages here and have the offsets committed to kafka?

duarten (commented):
LGTM in general!

src/consumer/base_consumer.rs
scanterog: That's a good question. My understanding is that messages won't be forwarded here anymore. In this particular case we had to switch [rd_kafka_consumer_close](https://github.com/confluentinc/librdkafka/blob/95a542c87c61d2c45b445f91c73dd5442eb04f3c/src/rdkafka.h#L4207-L4212) by [rd_kafka_consumer_close_queue](https://github.com/confluentinc/librdkafka/blob/95a542c87c61d2c45b445f91c73dd5442eb04f3c/src/rdkafka.h#L4228-L4235) because `rd_kafka_consumer_close` will block until the consumer has revoked its assignment by calling `Unassign` or `IncrementalUnassign` on the rebalance handler. This translates on the consumer being blocked forever as there's no rebalance callback registered anymore.

Hence we had to move to `rd_kafka_consumer_close_queue` that allows us to asynchronously close the consumer. In this particular case, by calling `Poll` we can fwd the rebalance event to the rebalance method on the `ConsumerContext` following the API doc:
> The application must poll/serve this queue until rd_kafka_consumer_closed() returns true.

src/consumer/base_consumer.rs
scanterog: It seems to work fine for the following common cases. Did I miss something?

- If the user provides a `timeout` set to 100ms, then the `op_timeout` will be 100ms (as `min_poll_interval` is 1sec by default). If no msgs or errors have been returned, then on this check `op_timeout` is equal to `timeout` and we return `None`.
- If the user provides a `timeout` set to 2 seconds (over the default for `min_poll_interval`), then `op_timeout` will be 1 second. Let's say we poll for 1 second and it does not return a msg/error, then `op_timeout` is less than timeout and now we do `timeout -= op_timeout` which set `timeout` to 1 second. If we get None messages/errors again, then `op_timeout` is equal to `timeout` and we return `None`.
- If the user provides `timeout` set to Never, then `op_timeout` is set to the default of 1 sec and we keep polling every second until we get a message or error.

duarten: I was concerned about something like the following:

* the user sets the `timeout` to 100ms. 
* After 1ms, we get `RD_KAFKA_EVENT_OFFSET_COMMIT` event. 
* We do `timeout -= op_timeout`, which puts `timeout` at 0. 
* We exit.

src/consumer/base_consumer.rs
duarten: Alright, sounds good to me.

src/consumer/base_consumer.rs
scanterog: ah that's a good catch. Ok, I'll fix this.

duarten (approved):

davidblewett (approved):
Looks great!

mensfeld:
@scanterog 
> > Is it not possible to yield from rust in the context of rebalance callback? How will it operate on rebalanes where the revocation happens first and prior to it an action like offset storage could take place?
> 
> Hello @mensfeld! The rebalance handler does not really change here for rust-rdkafka. Is there anything in particular that gives that impression? Can you elaborate a little bit more about this?

This:

it allows handling the rebalance events outside of the scope of the registered callback. This is specially useful for services using rust-rdkafka to build FFI libraries on top of it.

Since rebalance callbacks are triggered from the thread that is polling the consumer queue (one or both with main), a closure in the callback should allow to run code "outside" (in a way). So I wondered how this is done via events API.

I am myself looking to move to the IO event API for producer and admin but not sure what benefits it yields for regular high level consumers.

mensfeld:
@scanterog thank you for your explanation. That simplified things for me a lot :pray:

src/consumer/base_consumer.rs
Millione: Is there any chance that the consumer never closed? Because I ran into a situation where my program never exit this while loop.

Millione:
Is there any chance that the consumer never closed in BaseConsumer drop method? Because I ran into a situation where my program never exit that while loop checking whether closed.

mensfeld:
@Millione are you using cooperative sticky assignment strategy?

Millione:
> @Millione are you using cooperative sticky assignment strategy?

No, will it make a difference?

mensfeld:
@Millione I thought you ended up with an infinite loop error that can happen with certain combination of cooperative-sticky and some specific partition counts. Please ignore if not used because then it's not the case.

Millione:
@mensfeld Even if in the case you said, I still think it's not reasonable here to loop forever. The implementation should prevent this?

mensfeld:
@Millione I cannot tell you much about the implementation in rust because I maintain the implementation in Ruby. The issue here is, that polling delegates the responsibility to the C layer and it is the C layer that may hang. If that happens it happens. Maybe there are tools to exit out of such hanging code available in Rust, but at least in Ruby, they are of low quality and do not fix anything.

I do not know what is happening in your particular case, but I hold the Rust bindings in high regard in terms of stability and quality.

Millione:
@mensfeld Thanks for your patient explanation, I finally find I am stuck in the poll stage for rebalancing while closing, the last in stack frame is pthread cond wait.

PR #644, opened 2024-01-08
Return back to the caller on rebalance events

scanterog:
This does not affect the StreamConsumer or any other wrapper consumer. It will only incur on an extra Poll call when there's a rebalance event which is negligible. 

why is this needed?

When using bindings built upon the rust-rdkafka ffi, the caller is responsible for initiating the rebalance calls (*assign). If a high timeout is specified, the rebalance handler will only be triggered once the timeout period has elapsed. This fixes it by always returning on rebalance events.

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -144,9 +144,15 @@ where
                     }
                     rdsys::RD_KAFKA_EVENT_REBALANCE => {
                         self.handle_rebalance_event(event);
+                        if timeout != Timeout::Never {
+                            return None;
+                        }
                     }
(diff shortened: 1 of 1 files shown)


davidblewett (approved):
LGTM.

acautin:
@scanterog or @davidblewett We have a job that starts, check for messages / process then shuts down again as it receives messages quite sporadically (1 or 2 a day).

After this change: When we initialize the client there is a rebalance. The first call to `consumer.poll(Duration::from_secs(60))` returns `None` so the job terminates before reading any messages. Is this intended and if so what can be done to read all the messages ? 🤔

PR #655, opened 2024-02-14
fix: forward log messages

trtt:
Transition from callbacks to events changed log handling for librdkafka's internal threads - logs are sent to stderr (default in librdkafka).
This PR fixes it and restores previous forwarding behavior.


trtt:
https://github.com/fede1024/rust-rdkafka/issues/665 should be fixed first

PR #669, opened 2024-04-10
Poll error callbacks

ignytis:
Hello,

This PR adds a callback logic for errors occurred on `poll` method call in producer. The idea is to handle several kinds of errors immediately.

Here is a use case for this feature. If the application fails to connect or to authenticate to Kafka server then it exits with error code:
https://github.com/ignytis/send2kafka/blob/705f022beb50d8c8be77abbf6e15bd79707e9ebe/src/kafka_producer.rs#L32 <- setting the callback
https://github.com/ignytis/send2kafka/blob/705f022beb50d8c8be77abbf6e15bd79707e9ebe/src/kafka_producer.rs#L8-L13 <- callback itself
This way the connection issues might be handled before trying to send actual messages to the broker.

I'd appreciate any recommendations if there are any option to improve this code or implement this feature the other way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1070,7 +1070,7 @@ dependencies = [
 
 [[package]]
 name = "rdkafka"
-version = "0.36.2"
+version = "0.36.3"
 dependencies = [
  "async-std",
  "backoff",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rdkafka"
-version = "0.36.2"
+version = "0.36.3"
 authors = ["Federico Giraud <giraud.federico@gmail.com>"]
 repository = "https://github.com/fede1024/rust-rdkafka"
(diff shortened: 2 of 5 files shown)


rhishikeshj:
Perhaps a better design in keeping with the existing code would be to add an error callback to the ProducerContext trait similar to the `delivery` callback. And return some other useful error info. librdkafka also exposes functions for getting information about the errors, stuff like whether they are fatal and require termination etc. More is explained [here](https://github.com/confluentinc/librdkafka/blob/master/INTRODUCTION.md#threads-and-callbacks)
Since this is an old PR, maybe you have already moved on to something else, or your use-case is well covered by this implementation.

ignytis:
Looks like it's implemented in https://github.com/fede1024/rust-rdkafka/pull/771 indeed.

Anyway, thanks for pinging here, now I can close this old PR :)

PR #671, opened 2024-04-13
Use CStr::from_bytes_until_nul to read rdkafka config

chklauser:
Commit 353812ff958b4b65f9e65dd22a60e770ed6ba948 replaced the more finicky `CStr::from_bytes_with_nul` with `String::from_utf8_lossy`. This causes tests to fail, both for me locally and in [CI](https://github.com/fede1024/rust-rdkafka/actions/runs/8417735908/job/23046767982). 

That commit eliminated a panic but now the returned Rust string (from `NativeClientConfig.get`) contains NUL bytes. This might have accidentally worked in FFI APIs that take C strings, but it fails when such a string is passed to a Rust API, such as integer parsing in `stream_consumer.rs:217`.

The newer `CStr::from_bytes_until_nul` function
 1. does not panic
 2. ends the string at the first NUL byte

It does return an error when there is no NUL byte in the input slice.

While it would be possible to fall back to `String::from_utf8_lossy` in that case, I'm not sure that would be a good idea. If we are in that situation, librdkafka clearly has messed something up completely. The contents of the buffer are then more likely complete garbage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
     branches: [master]
 
 env:
-  rust_version: 1.61.0
+  rust_version: 1.69.0
 
 jobs:
   lint:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["kafka", "rdkafka"]
 categories = ["api-bindings"]
 edition = "2018"
 exclude = ["Cargo.lock"]
-rust-version = "1.61"
+rust-version = "1.69"
 
 [workspace]
 members = ["rdkafka-sys"]
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ re-exported as rdkafka features.
 
 ### Minimum supported Rust version (MSRV)
 
-The current minimum supported Rust version (MSRV) is 1.61.0. Note that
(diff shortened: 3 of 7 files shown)


chklauser:
https://github.com/fede1024/rust-rdkafka/commit/353812ff958b4b65f9e65dd22a60e770ed6ba948 has not been released on crates.io as far as I can tell. => No need to add the bugfix to the changelog.

Fenex:
Important note: `CStr::from_bytes_until_nul` required update MSRV to `1.69.0` (current is `1.61.0`).

chklauser:
Indeed! I've included a bump to version 1.69 in the PR (as a separate commit). 

Rust 1.69 was released 20 April, 2023. Seems a reasonable upgrade to me, but I don't really have a stake in that decision (we are using latest stable).

chklauser:
It seems there are a total of three PRs for this error now :( 

1. https://github.com/fede1024/rust-rdkafka/pull/671 (this PR)
2. https://github.com/fede1024/rust-rdkafka/pull/688
3. https://github.com/fede1024/rust-rdkafka/pull/691

mati865:
There is one more PR with the fix but contains also other changes: https://github.com/fede1024/rust-rdkafka/pull/647

chklauser:
Closed in favor of #688

PR #680, opened 2024-05-05
Fix panic in GroupInfo::members

jeremija:
I recently observed undefined behavior in one of the projects I'm working on because [slice::from_raw_parts] is called without a null-check in `GroupInfo::members`. This undefined behavior was present when iterating over the resulting slice and it would just terminate prematurely when trying to chain multiple iterators. The function is pretty strict about what kind of pointers it accepts:

> data must be non-null and aligned even for zero-length slices.

This undefined behavior has become a panic in debug builds in [Rust 1.78.0]:

> For example, slice::from_raw_parts requires an aligned non-null pointer.
> The following use of a purposely-misaligned pointer has undefined behavior,
> and while if you were unlucky it may have appeared to "work" in the past,
> the debug assertion can now catch it:

Cause is found in [rdkafka.c]. I see there are more uses of `slice::from_raw_parts` in the code without the NULL checks, but it is tricky for me to figure if any of those pointers can be NULL, and whether any of those calls should be replaced with `ptr_to_slice`.

[slice::from_raw_parts]: https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html
[Rust 1.78.0]: https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#asserting-unsafe-preconditions
[rdkafka.c]: https://github.com/confluentinc/librdkafka/blob/95a542c87c61d2c45b445f91c73dd5442eb04f3c/src/rdkafka.c#L4668-L4670


davidblewett:
@jeremija can you fix the conflicts reported here?

PR #688, opened 2024-05-29
Remove nul chars from string before parsing to numerical values

KowalczykBartek:
Looks like this change https://github.com/fede1024/rust-rdkafka/commit/353812ff958b4b65f9e65dd22a60e770ed6ba948 is causing following error (when SimpleConsumer is being run) - at least on OSX 
```
librdkafka validated config value is valid u64: ParseIntError { kind: InvalidDigit }
```
That is because rd_kafka_conf_get returns null terminated string, 
```rust
// Allocate a buffer of that size and call again to get the actual
// string.
let mut buf = vec![0_u8; size];
let res = unsafe {
    rdsys::rd_kafka_conf_get(
        self.ptr(),
        key_c.as_ptr(),
        buf.as_mut_ptr() as *mut c_char,
        &mut size,
    )
};
```
ends with buf equal:
```
[51, 48, 48, 48, 48, 48, 0];
```
I guess reverting this change will cause issues on windows, so maybe my PR will be good enough to handle this.

This pr should also fix this case https://github.com/fede1024/rust-rdkafka/pull/683

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -150,7 +150,7 @@ impl NativeClientConfig {
         }
 
         // Convert the C string to a Rust string.
-        Ok(String::from_utf8_lossy(&buf).to_string())
+        Ok(String::from_utf8_lossy(&buf).trim_matches(char::from(0)).to_string())
     }
 }
 


chklauser:
Just as a heads up for maintainers: there seem to be 3 PRs for this issue

1. https://github.com/fede1024/rust-rdkafka/pull/671
2. https://github.com/fede1024/rust-rdkafka/pull/688  (this PR)
3. https://github.com/fede1024/rust-rdkafka/pull/691

fede1024 (approved):

mati865:
This PR feels like a workaround that is doomed to break in the future. Aby chance to revisit https://github.com/fede1024/rust-rdkafka/pull/671?

fede1024:
I agree that #671 is a better solution. @mati865 any chance you could file a PR with `from_bytes_until_nul`?  #671 has a few other changes that are conflicting with the master branch. Thanks

mati865:
I'll wait for @chklauser to reply if he wants to review his PR, otherwise I'll rebase hit work.

chklauser:
I have rebased the change as #705 (forgot to re-open _before_ pushing the new branch). With the unrelated bump of MSRV to 1.70 we can now just use `CStr::from_bytes_until_nul` 🥳 

Thanks for tagging me @mati865

andrei-ionescu:
I also created #706 for this issue, not knowing that there is this one. If we can have any of the two merged would be great.

PR #689, opened 2024-06-02
Enable the development of custom consumers

manifest:
This PR for the [issue](https://github.com/fede1024/rust-rdkafka/issues/690).

### DIFF
--- a/src/consumer/base_consumer.rs
+++ b/src/consumer/base_consumer.rs
@@ -38,6 +38,7 @@ where
     client: Client<C>,
     queue: NativeQueue,
     group_id: Option<String>,
+    nonempty_callback: Option<Box<Box<dyn Fn() + Send + Sync>>>,
 }
 
 impl FromClientConfig for BaseConsumer {
@@ -98,6 +99,7 @@ where
             client,
             queue,
             group_id,
+            nonempty_callback: None,
         })
     }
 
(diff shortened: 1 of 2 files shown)


fede1024:
Thank you for the contribution. Can you please add a test to verify that the callback is getting called. Thanks

manifest:
Sure, I will be able to look into it soon.

Do you have any suggestion on how I can trigger `native_message_queue_nonempty_cb` callback in the test?

manifest:
I've added the test.

fede1024:
Thank you for adding the test. Can we produce a real message instead of calling `rd_kafka_queue_yield`? Tests are supposed to run with a working Kafka deployment alongside it (via docker compose), so you should be able to produce and consume from real topics.

manifest:
I've moved the test to the `tests/test_low_consumers` module and replicated scenarios from [`test_produce_consume_message_queue_nonempty_callback`](https://github.com/fede1024/rust-rdkafka/blob/master/tests/test_low_consumers.rs#L356-L448).

Statistics is [turned off](https://github.com/fede1024/rust-rdkafka/pull/689/files#diff-90afe39ccc4f1169473bb128acc2565c145a21c5e4f845ec5c4a3bdf3b874f06R458-R460) to prevent interference with the wakeups counter.

manifest:
The failure of the test on the 7.5.1 + 3.5 option doesn't indicate an issue with the changes in the current PR, as the test also fails on the existing `test_produce_consume_message_queue_nonempty_callback` test.

PR #707, opened 2024-08-07
Force auth when fetching Kafka metadata

andrei-ionescu:
Kafka metadata may also require authentication. This PR forces authentication before fetching Kafka metadata.


andrei-ionescu:
@fede1024, @benesch, @davidblewett Can some one have a look into this small change?

andrei-ionescu:
@fede1024, @benesch, @davidblewett Can you have a look into this minor change? If this needs other things let me know.

PR #712, opened 2024-08-21
Avoid realloc when creating C-style strings

mlowicki:
Pass ref to benefit from underlying specialization - https://doc.rust-lang.org/1.80.1/src/alloc/ffi/c_str.rs.html#279


davidblewett (approved):
This looks good to me. The `test_produce_consume_consumer_nonempty_callback` has been flaky.

Issue #614, opened 2023-10-15
Add support for AdminAPI `DescribeCluster()` and `DescribeTopics()`


dkirrane:
I'm looking for the rust equivalent of the Java `org.apache.kafka.clients.admin.AdminClient` API

```
            adminClient
                .describeCluster(
                    new DescribeClusterOptions()
                        .timeoutMs((int) Math.min(Integer.MAX_VALUE, remainingWaitMs)))
                .nodes()
                .get();
``

and

```
        adminClient.describeTopics(
            topicNames, new DescribeTopicsOptions().timeoutMs(Math.toIntExact(timeoutMs)));
```

It looks like it's available in librdkafka now via this commit
https://github.com/confluentinc/librdkafka/pull/4240/files

PR #721, opened 2024-09-13
Add DescribeTopics support for admin client API

alihan-synnada:
Adding DescribeTopics now and also potentially DescribeCluster in this PR as per https://github.com/fede1024/rust-rdkafka/issues/614

I'll write the docs and tests, and then mark as ready for review

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -28,6 +28,7 @@ use crate::log::{trace, warn};
 use crate::util::{cstr_to_owned, AsCArray, ErrBuf, IntoOpaque, KafkaDrop, NativePtr, Timeout};
 use crate::TopicPartitionList;
 
+use uuid::Uuid;
 //
 // ********** ADMIN CLIENT **********
 //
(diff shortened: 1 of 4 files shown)


src/admin.rs
```diff
@@ -1340,3 +1378,184 @@ impl Future for AlterConfigsFuture {
...
+        Poll::Ready(Ok(out))
+    }
+}
+
+fn extract_topic_id(topic_description: *const RDKafkaTopicDescription) -> Uuid {
+    let topic_id = unsafe { rdsys::rd_kafka_TopicDescription_topic_id(topic_description) };
+    let high_bits = unsafe { rdsys::rd_kafka_Uuid_most_significant_bits(topic_id) } as u64;
+    let low_bits = unsafe { rdsys::rd_kafka_Uuid_least_significant_bits(topic_id) } as u64;
+    Uuid::from_u64_pair(high_bits, low_bits)
+}
```
alihan-synnada, line 1477: Is it really worth adding a dependency to the `uuid` crate when `Uuid` is only used here? We might as well remove the dependency and only return a `(high_bits, low_bits)` tuple.

alihan-synnada:
I added DescribeCluster support as well. DescribeConsumerGroups should also be relatively easier to implement since the some of the types it returns (`Node` and `AclOperation`) are implemented now.

alihan-synnada:
I'm satisfied with the PR myself but there are a few points in my mind:
- I'm not sure if the tests are covering enough ground.
- I couldn't decide where to place structs and enums so they are somewhat all over the place.
- Should we add the `uuid` crate as a dependency? The `Uuid` type is more descriptive and the `uuid` crate is rather lean when no features are enabled. On the other hand, it's only used for `topic_id` in `TopicDescription` so we can remove the dependency and simply use `(u64, u64)` or `u128` instead.

alihan-synnada:
@davidblewett @benesch Any opportunities to review and possibly merge within next week?

gzsombor:
Can someone please review and merge this feature? I've rebased it in my [fork](https://github.com/gzsombor/rust-rdkafka/tree/feature/describe-topics) and seems to be working fine, but love to avoid using a git dependency. 
Thank you!

PR #722, opened 2024-09-14
Update MSRV to 1.70

Abdullahsab3:
I think https://github.com/fede1024/rust-rdkafka/pull/703 updated Rust's MSRV to 1.70. It might be good to update the Readme accordingly as well.

Closes https://github.com/fede1024/rust-rdkafka/issues/720

PR #724, opened 2024-09-18
sys: Upgrade num_enum to 0.7

svix-jplatte:
Currently, rdkafka pulls in `syn 1.0` via `num_enum 0.5`. Most parts of the Rust ecosystem have moved to `syn 2.0`, and it's annoying to have both in the dependency tree because it's pretty heavy in terms of compile time.

I chose to first update `Cargo.lock` via `cargo update` because it was very outdated, and to show that the actual lockfile diff for the `num_enum` upgrade you would also see elsewhere from this doesn't pull in anything extra. I can update the PR to only do the minimal lockfile upgrade if you prefer.

I also checked for breaking changes. `num_enum` unfortunately doesn't have a changelog, but it seems like the only relevant changes are:

- num_enum 0.6 changed some behavior w.r.t. `#[default]` / `#[num_enum(default)]` attributes, which you aren't using
- num_enum 0.7 increases the minimum supported Rust version to 1.70.0, which matches the current MSRV here as of #722

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,36 +3,51 @@
 version = 3
 
 [[package]]
-name = "adler32"
-version = "1.0.4"
+name = "addr2line"
+version = "0.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5d2e7343e7fc9de883d1b0341e0b13970f764c14101234857d2ddafa1cb1cac2"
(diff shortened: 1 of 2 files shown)


svix-jplatte:
Gentle ping @benesch @davidblewett (you seem to have recently reviewed other PRs)

svix-jplatte:
CI results mostly match `main`, though `(7.5.1, 3.6)` succeeded there. I'm presuming `test_produce_consume_consumer_nonempty_callback` is flaky?

PR #731, opened 2024-10-03
make `ThreadedProducer` clonable like `BaseProducer`

petrosagg:
The threaded producer was not `Clone` due to handling the background thread join handle but that can easily be stored in an `Arc` that the last drop of the client unwraps and terminates the background thread.

`Arc::into_inner` guarantees that exactly one thread will manage to unwrap the handle.


benesch (approved):
Thank you!

PR #748, opened 2024-12-05
Various BaseProducer::flush fixes

davidblewett:
Refactor `BaseProducer::poll` to not return early, but instead continue processing events until the passed timeout.

Refactor `BaseProducer::flush` to spend most time in `librdkafka`, and whatever is left in `BaseProducer::poll`.

### DIFF
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -366,7 +366,7 @@ where
             Duration::MAX
         };
         let mut attempt = 0;
-        while attempt > 0 && remaining > Duration::ZERO {
+        while attempt >= 0 && remaining > Duration::ZERO {
             let start = Instant::now();
             let event = self
                 .client()
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -361,8 +361,8 @@ where
     /// the message delivery callbacks.
     pub fn poll<T: Into<Timeout>>(&self, timeout: T) {
         let deadline: Deadline = timeout.into().into();
-        let mut attempt = 0;
-        while attempt >= 0 && !deadline.elapsed() {
+        let mut called = false;
+        while !called || !deadline.elapsed() {
             let event = self
                 .client()
                 .poll_event(&self.queue, Timeout::After(deadline.remaining()));
@@ -379,7 +379,7 @@ where
                     }
                 }
             }
-            attempt += 1;
+            called = true;
         }
     }
 
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -505,6 +505,8 @@ where
             };
             if ret.is_error() {
                 return Err(KafkaError::Flush(ret.into()));
+            } else if let Deadline::Never = &deadline {
+                self.poll(Timeout::After(Duration::ZERO));
             } else {
                 self.poll(&deadline);
             }
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -503,13 +503,14 @@ where
             let ret = unsafe {
                 rdsys::rd_kafka_flush(self.client().native_ptr(), deadline.remaining_millis_i32())
             };
-            if ret.is_error() {
-                return Err(KafkaError::Flush(ret.into()));
-            } else if let Deadline::Never = &deadline {
+            if let Deadline::Never = &deadline {
                 self.poll(Timeout::After(Duration::ZERO));
             } else {
                 self.poll(&deadline);
             }
+            if ret.is_error() {
+                return Err(KafkaError::Flush(ret.into()));
+            };
         }
         Ok(())
     }
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -508,12 +508,12 @@ where
         let mut remaining = if let Timeout::After(dur) = timeout.into() {
             dur
         } else {
-            // librdkafka's flush api requires an i32 millisecond timeout
-            Duration::from_millis(i32::MAX as u64)
+            Duration::MAX
         };
         while self.in_flight_count() > 0 && remaining > Duration::ZERO {
             let flush_start = Instant::now();
             let ret = unsafe {
+                // This cast to i32 will truncate to i32::MAX
                 rdsys::rd_kafka_flush(self.client().native_ptr(), remaining.as_millis() as i32)
             };
             if ret.is_error() {
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -360,11 +360,7 @@ where
     /// Regular calls to `poll` are required to process the events and execute
     /// the message delivery callbacks.
     pub fn poll<T: Into<Timeout>>(&self, timeout: T) {
-        let deadline = if let Timeout::After(dur) = timeout.into() {
-            Deadline::new(dur)
-        } else {
-            Deadline::new(Duration::MAX)
-        };
+        let deadline: Deadline = timeout.into().into();
         let mut attempt = 0;
         while attempt >= 0 && !deadline.elapsed() {
             let event = self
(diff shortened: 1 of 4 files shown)


src/producer/base_producer.rs
```diff
@@ -494,26 +505,27 @@ where
...
     // the librdkafka-handled message count to reach zero. Runs until value reaches zero or timeout.
     fn flush<T: Into<Timeout>>(&self, timeout: T) -> KafkaResult<()> {
-        let mut timeout = timeout.into();
-        loop {
-            let op_timeout = std::cmp::min(timeout, self.min_poll_interval);
-            if self.in_flight_count() > 0 {
-                unsafe { rdsys::rd_kafka_flush(self.native_ptr(), 0) };
-                self.poll(op_timeout);
+        let mut remaining = if let Timeout::After(dur) = timeout.into() {
+            dur
```
mlowicki, line 509: if API requires i32 then we need to check it here too. What if `dur` will be e.g. `i32::MAX + 1` ?

src/producer/base_producer.rs
```diff
@@ -362,19 +360,32 @@ where
...
-                        let evname = rdsys::rd_kafka_event_name(ev.ptr());
-                        CStr::from_ptr(evname).to_string_lossy()
-                    };
-                    warn!("Ignored event '{}' on base producer poll", evname);
+        let mut remaining = if let Timeout::After(dur) = timeout.into() {
+            dur
+        } else {
+            Duration::MAX
+        };
+        let mut attempt = 0;
```
mlowicki, line 368: if `attempt` is 0 then we won't run even single iteration of the loop below.

src/producer/base_producer.rs
```diff
@@ -362,19 +360,32 @@ where
...
+        while attempt > 0 && remaining > Duration::ZERO {
+            let start = Instant::now();
```
mlowicki, line 370: You can simplify code here by introducing simple utility type for measuring  time.

```
struct Deadline {
    deadline: Instant
}

impl Deadline {
    fn new(duration: Duration) -> Self {
         Self { deadline: Instant::now().add(...
    }

   fn remaining(&self) -> Duration {
      ...
   }
 
  fn elapsed() -> bool { 
      ....
  }
}
```

src/producer/base_producer.rs
davidblewett: Fixed in 3f2b5b91.

src/producer/base_producer.rs
davidblewett: Fixed in 204dbe2f.

src/producer/base_producer.rs
davidblewett: Fixed in 36d4ebfa.

src/producer/base_producer.rs
mlowicki: Do we need `attempt` at all? It starts with 0 as it only incremented (so always >= 0). In while loop condition `attempt >= 0` so it'll be always true.

src/producer/base_producer.rs
```diff
@@ -494,26 +503,26 @@ where
...
-                unsafe { rdsys::rd_kafka_flush(self.native_ptr(), 0) };
-                self.poll(op_timeout);
+        let deadline = if let Timeout::After(dur) = timeout.into() {
+            Deadline::new(dur)
+        } else {
+            Deadline::new(Duration::MAX)
+        };
+        while self.in_flight_count() > 0 && !deadline.elapsed() {
+            let ret = unsafe {
+                // This cast to i32 will truncate to i32::MAX
```
mlowicki, line 513: I'm not sure about it. `Duration::as_millis()` return `u128` (https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_millis). So I think if you cast something like i32::MAX + 1 then you'll get negative.

```rust
use std::time::Duration;

fn main() {
    let d = Duration::from_millis(2_147_483_647 /* i32::Max */ + 1);
    println!("{}", d.as_millis() as i32);
}
```

output: -2147483648


https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f2cda565c0c554ef158e5ea3c9a17921

davidblewett: Wow; that's really surprising. I will update and handle explicitly.

src/producer/base_producer.rs
davidblewett: I added that to ensure that if someone passed `Timeout::After(Duration::ZERO)` for a non-blocking call, we would always call the underlying `librdkafka` method at least once.

mlowicki: not really as `&&` is used in the condition.

davidblewett: Okay; I believe I fixed the logic error in 1957d2f01249ad96c887c4f365f6cd26ef1394a8.

src/producer/base_producer.rs
```diff
@@ -362,19 +360,30 @@ where
...
-            let evtype = unsafe { rdsys::rd_kafka_event_type(ev.ptr()) };
-            match evtype {
-                rdsys::RD_KAFKA_EVENT_DR => self.handle_delivery_report_event(ev),
-                _ => {
-                    let evname = unsafe {
-                        let evname = rdsys::rd_kafka_event_name(ev.ptr());
-                        CStr::from_ptr(evname).to_string_lossy()
-                    };
-                    warn!("Ignored event '{}' on base producer poll", evname);
+        let deadline = if let Timeout::After(dur) = timeout.into() {
```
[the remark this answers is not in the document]
davidblewett: Added the impl in 25951bf5eb278449a341afc22b21292d7751ad31. This means we have to call `into()` twice, but does centralize the logic.

src/producer/base_producer.rs
```diff
@@ -362,19 +360,26 @@ where
...
-                rdsys::RD_KAFKA_EVENT_DR => self.handle_delivery_report_event(ev),
-                _ => {
-                    let evname = unsafe {
-                        let evname = rdsys::rd_kafka_event_name(ev.ptr());
-                        CStr::from_ptr(evname).to_string_lossy()
-                    };
-                    warn!("Ignored event '{}' on base producer poll", evname);
+        let deadline: Deadline = timeout.into().into();
+        let mut called = false;
+        while !called || !deadline.elapsed() {
```
mlowicki, line 365: nit: more idiomatic way would be:

```
loop {
    do_stuff();
    if !condition {
        break;
    }
}
```

src/util.rs
```diff
@@ -76,6 +107,36 @@ impl std::ops::SubAssign for Timeout {
...
+            Deadline::new(Some(cmp::min(Deadline::MAX_FLUSH_DURATION, dur)))
+        } else {
+            Deadline::new(None)
+        }
+    }
+}
+
+impl From<Deadline> for Timeout {
+    fn from(d: Deadline) -> Timeout {
+        if let Deadline::Never = d {
```
mlowicki, line 122: maybe we could use `impl From<&Deadline> for Timeout` here to not duplicate the logic?

src/producer/base_producer.rs
```diff
@@ -494,26 +499,22 @@ where
...
-            if self.in_flight_count() > 0 {
-                unsafe { rdsys::rd_kafka_flush(self.native_ptr(), 0) };
-                self.poll(op_timeout);
+        let deadline: Deadline = timeout.into().into();
+        while self.in_flight_count() > 0 && !deadline.elapsed() {
+            let ret = unsafe {
+                // This cast to i32 will truncate to i32::MAX
+                rdsys::rd_kafka_flush(
+                    self.client().native_ptr(),
+                    deadline.remaining().as_millis() as i32,
```
mlowicki, line 508: maybe `Deadline` could have a dedicated method instead like `remaining_millis_i32` and inside it we could deal with everything related to conversion? 

We could avoid having https://github.com/fede1024/rust-rdkafka/pull/748/files#diff-4007187965c1fea4fd252a76f5f1007b2ea59f6d86e2f454edee547c5036be24R113 and inside `remaining_millis_i32` do the necessary clamping.

Currently it's clunky as we assume here that `remaining()` or `Deadline` in general is doing the clamping under the hood.

mlowicki (approved):

src/producer/base_producer.rs
davidblewett: Fixed in 00e64ed1.

src/producer/base_producer.rs
davidblewett: Fixed in 68ed9061.

src/util.rs
```diff
@@ -76,6 +111,26 @@ impl std::ops::SubAssign for Timeout {
     }
 }
 
+impl From<Timeout> for Deadline {
+    fn from(t: Timeout) -> Deadline {
+        if let Timeout::After(dur) = t {
+            Deadline::new(Some(cmp::min(Deadline::MAX_FLUSH_DURATION, dur)))
```
mlowicki, line 117: this shouldn't be needed anymore as we've `remaining_millis_i32()` now.

davidblewett: Fixed in ab7b4af21590408e4ba3562994312e9889a1b934.

PR #750, opened 2024-12-06
add feature to statically link against existing librdkafka

aviramha:
This feature allows users to pre-build their own static librdkafka and use it with rust-rdkafka.
In our case, to make it easier for cross compilation we preferred to compile natively librdkafka for each architecture, then re-use it from a container image.


aviramha:
For those interested, we started maintaining our own fork until this is merged
https://github.com/metalbear-co/rust-rdkafka

fede1024:
TY!

aviramha:
Thanks for merging!

PR #752, opened 2024-12-13
README.dm: Update timely-dataflow link to the current repository.

davideagen:
The originally linked repository hasn't been updated in 5 years and is a fork of the active repository, which is https://github.com/TimelyDataflow/timely-dataflow

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ logging framework.
 [rdkafka-sys-known-issues]: https://github.com/fede1024/rust-rdkafka/tree/master/rdkafka-sys/README.md#known-issues
 [smol]: https://docs.rs/smol
 [timely-blog]: https://github.com/frankmcsherry/blog/blob/master/posts/2017-11-08.md
-[timely-dataflow]: https://github.com/frankmcsherry/timely-dataflow
+[timely-dataflow]: https://github.com/TimelyDataflow/timely-dataflow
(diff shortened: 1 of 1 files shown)


benesch (approved):

davideagen:
Thanks for the quick response, but I just realized that it really should still link to the old repository. I didn't realize until after I made the PR that the [kafkaesque](https://github.com/frankmcsherry/timely-dataflow/tree/master/kafkaesque) subproject in the original repo linked from the [blog post](https://github.com/frankmcsherry/blog/blob/master/posts/2017-11-08.md) is **not** included in the active timely-dataflow repository.

So, I think my PR should be reverted.

benesch:
It was only removed earlier this year: https://github.com/TimelyDataflow/timely-dataflow/pull/545

So one option would be to link to an older version of timely. The other option is just to strike the link entirely since it's no longer included in the latest version of timely.

davideagen:
Perhaps it is best to remove the link entirely, given that it was removed from the main timely repo. With both you and Frank at Materialize, according to your GitHub profiles, I'll let you two decide what's best. If you want to keep the links, I've created #753 to revert the change.

benesch:
Yeah, after consideration, let's just strike it from the README. I updated #753 to do so and merged. Thanks!

PR #759, opened 2025-02-13
Detailed logs for base producer poll error events

EddieWhi:
Hi Maintainers,

We have a number of mystery logs in our systems using rust-rdkafka along the lines of:

```Ignored event 'Error' on base producer poll```

I created this simple branch to get the error details out so we could diagnose and fix. The issue was fairly simple in the end (connections were timing our through AWS LBs, recommended settings for `connections.max.idle.ms` were missing). It would be great if I could get this (or similar) into a release someday to help us diagnose this kind of issue in the future.

I can see that this change could subtly impact current users as new error events will be pushed through `ClientContext::error`. If that's considered acceptable and manageable then great. Alternatively could have pushed the error detail through `ClientContext::log` which feels like it's less likely to cause issue although it's a little unsymmetrical?

Any thoughts and help getting this in would be greatly appreciated.

### DIFF
--- a/src/producer/base_producer.rs
+++ b/src/producer/base_producer.rs
@@ -367,11 +367,13 @@ where
                 let evtype = unsafe { rdsys::rd_kafka_event_type(ev.ptr()) };
                 match evtype {
                     rdsys::RD_KAFKA_EVENT_DR => self.handle_delivery_report_event(ev),
+                    rdsys::RD_KAFKA_EVENT_ERROR => self.handle_error_event(ev),
                     _ => {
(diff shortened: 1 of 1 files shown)


rhishikeshj:
Would this also be a good place to discuss and perhaps add support for error event callback in the ProducerContext trait? 
Knowing about producer side errors specially ones which are categorised as [fatal](https://github.com/confluentinc/librdkafka/blob/master/INTRODUCTION.md#threads-and-callbacks) would be helpful

PR #766, opened 2025-04-02
Require a minimum of CMake 3.5 compatibility for librdkafka build.

wesc:
The most recent CMake (released March 28, 2025) drops 3.2 compatibility, which the version of librdkafka in rdkafka-sys (2.3.0) depends on. Recent versions of librdkafka increase the min cmake version to 3.5. Rather than upgrade rdkafka-sys to a new librdkafka or patch that source, this PR simply specifies a new min CMake version in the `build.rs` file.

To check that it's working, you can just run `cargo build --features cmake-build`, which fails on head in master.

### DIFF
--- a/rdkafka-sys/build.rs
+++ b/rdkafka-sys/build.rs
@@ -213,7 +213,10 @@ fn build_librdkafka() {
         // want a stable location that we can add to the linker search path.
         // Since we're not actually installing to /usr or /usr/local, there's no
         // harm to always using "lib" here.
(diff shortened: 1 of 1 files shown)


vkill:
P.S. `Compatibility with versions of CMake older than 3.5 has been removed.` in CMake 4.0
https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features

fede1024:
Release 0.38.0 has been published.

PR #784, opened 2025-08-05
PartitionEOF: carry topic + offset so consumers can detect full catch-up

florin-akermann:
Right now `KafkaError::PartitionEOF` only hands you the partition number as an `i32`. Once a consumer is subscribed to more than one topic that's ambiguous — you get an EOF for partition 3, but nothing tells you which topic it belongs to or the offset it stopped at. This PR changes the variant to carry a small `TopicPartitionOffset` struct (topic, partition, offset).

Motivation: with the full coordinates you can track whether you've reached EOF on every partition you care about. With just the partition number you can't, since there's no way to tie an EOF back to a topic.

A concrete example is an app that builds state off a set of Kafka topics. It can watch for EOF across all its partitions and only flip to "ready" once it's caught up — which is handy for blue/green rollouts, where a new instance warms its state before it starts taking traffic. As a bonus the offset at EOF is the log-end position, so you also learn where you caught up to.

None of this is strictly new: you can already get the same signal by comparing your position to the high watermark via `fetch_watermarks`. This just makes the event-driven version pleasant — no polling loop, no watermark-vs-consume race. EOF is a one-shot signal, so it's better suited to an initial warm-up gate than a continuous health check.

It's a breaking change to `PartitionEOF`. If you'd rather keep it additive I'm happy to rework it.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,7 +4,12 @@ See also the [rdkafka-sys changelog](rdkafka-sys/changelog.md).
 
 ## Unreleased
 
-None
+* **Breaking change:** `KafkaError::PartitionEOF` now carries a
+  `TopicPartitionOffset` (topic, partition, and offset) instead of only the
+  partition `i32`, providing full context when a partition reaches EOF
+  ([#784]).
+
+[#784]: https://github.com/fede1024/rust-rdkafka/pull/784
 
 ## 0.39.0 (2026-01-25)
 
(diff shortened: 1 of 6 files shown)


florin-akermann:
Hi @fede1024 any chance we can get this merged?

florin-akermann:
Gave this a rebase and a cleanup. It's green locally on the pinned 1.85 toolchain — fmt, clippy `-Dwarnings` (lib and tests), `build --all-targets`, and a new integration test (`test_partition_eof_error_details`) that runs against a real broker via testcontainers.

Would appreciate a look when you get a chance. It's a breaking change to `PartitionEOF`, but I'm glad to make it additive if you'd rather. @davidblewett @fede1024 — any steer on the API direction?

PR #785, opened 2025-08-11
Additional AdminClient operations.

j-santander:
This PR adds three admin operations that were not previously supported:
- AdminClient::list_offsets: List offsets for the specified  topic_partitions. 
- AdminClient::describe_consumer_groups: Provides a description on the  requested consumer groups. 
- AdminClient::list_consumer_group_offsets: List offset information for the topics subscribed by a consumer group.
  
Implementation notes:
- This mostly my first Rust code, so please comment and critique. The work started because I was in search of something to do after having  read rust tutorials and documents. I decided for re-doing a python tool  I've recently coded for Kafka administration tasks and I found that someof the operations I was using in python were missing from the admin 
interface in rdkafka. Although the implementation I did in my rust tool  was different, I decided to try my hand at enhancing rdkafka and this PR is the result.
- For all three operations I've used a similar pattern for the result: `KafkaResult<Vec<Result<TYPE, (KEY, KafkaError)>`. This is similar to  the existing GroupResult. The alternative was to insert the error value in the TYPE record, following closer the librdkafka structure, but I felt it was less in line with rust conventions.
- I felt that I was adding many types and that it was best to define them in a different module/file for each operation. What I've done is re-exporting them from the admin module.
- I've defined types::RDKafkaXXX for all the bindings::rd_kafka_xxxx types, using the RDKafkaXXX alias in the code, keeping the function calls to rdsys::rd_kafka_xxxx
- Regarding GroupResult and ConsumerGroupResult: Technically they both come from the same rd_kafka_group_result_t, GroupResul just extracting the consumer group name. The GroupResult type was already used by the delete_groups operation. In order not to break the  existing API, I'm maintaining both. Perhaps in the future they could be enhanced (or perhaps it is well as it is, as the additional information is not meaningful in the delete_groups case).
- Added use cases covering the different operations. I've probably not been exhaustive, but I hope the coverage is enough.
- Documentation and field and structures descriptions I've taken from a mixture of the java, python and the protocol documentation. Hope it is in line with the rest of the documentation provided.

### DIFF
--- a/rdkafka-sys/src/types.rs
+++ b/rdkafka-sys/src/types.rs
@@ -96,6 +96,39 @@ pub type RDKafkaTopicResult = bindings::rd_kafka_topic_result_t;
 /// Native rdkafka group result.
 pub type RDKafkaGroupResult = bindings::rd_kafka_group_result_t;
 
+/// Native rdkafka list offsets result.
+pub type RDKafkaListOffsetsResult = bindings::rd_kafka_ListOffsets_result_t;
+
+/// Native rdkafka list offsets info returned.
+pub type RDKafkaListOffsetsResultInfo = bindings::rd_kafka_ListOffsetsResultInfo_t;
+
+/// Native result of the Describe Consumer Group operation.
+pub type RDKafkaDescribeConsumerGroupsResult = bindings::rd_kafka_DescribeConsumerGroups_result_t;
+
+/// Native description of a Consumer Group.
+pub type RDKafkaConsumerGroupDescription = bindings::rd_kafka_ConsumerGroupDescription_t;
+
+/// Native ACL Operation enum.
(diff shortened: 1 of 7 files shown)


Mcdostone:
Same, I would be interested to have this feature.

x10an14-nav:
Could you rebase this PR on top of master?

x10an14-nav:
> Could you rebase this PR on top of master?

I did it myself here: https://github.com/x10an14-nav/rust-rdkafka/tree/master [^1]
Seems to work: https://github.com/nais/klag-exporter/commit/cb6275f76a70211588a13b44cf8739510f49151f

1: Found 1x bug fixed in its latest commit (use after free)

andrei-ionescu:
I'm interested in this also

PR #787, opened 2025-08-25
Fix name of Cargo feature for static linking

cstyles:
d3879c2f29f7580da619e795b5e4883efbe5b6a8 (PR #750) added the ability to statically link to a pre-built version of librdkafka when the `static-linking` Cargo feature is enabled but the build script for rdkafka-sys checks for `static-external` instead. These need to match for static linking to take effect so let's fix the build script to use `static-linking`.


koh-satoh-wpg:
I would love to see this merged.

Issue #763, opened 2025-03-26
ClientConfig’s Debug implementation can lead to security issues


dsabadie-datadog:
Hello,

`ClientConfig` implements `Debug`, in which things such as credentials can be put in, such as `sasl.username`, `sasl.password`, etc.
It would be valuable to manually implement `Debug` and redact/obfuscate those to prevent leaking them publicly.

PR #793, opened 2025-09-28
Sanitizing ClientConfig Debug and Implementing basic Debug for AdminClient

j7nw4r:
Sanitizing the ClientConfig Debug to satisfy: https://github.com/fede1024/rust-rdkafka/issues/763

Implementing a basic AdminClient Debug so that KafkaResult<AdminClient<C>> can call unwrap_err(). Unwrap_err<T, E> requires that the success type T implements Debug. https://doc.rust-lang.org/core/result/enum.Result.html#method.unwrap_err

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -6,6 +6,7 @@
 
 use std::collections::HashMap;
 use std::ffi::{c_void, CStr, CString};
+use std::fmt;
 use std::future::Future;
 use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


j7nw4r:
I think I'm just going to work on a fork. The commit history on main (and what seems like maintainer burnout/inattentiveness) doesn't really give me confidence that I'll actually be able to contribute.

fede1024:
If you reopen the PR I would be happy to review it.

PR #798, opened 2025-10-02
test: converting test suite to TestContainers

j7nw4r:
Migrate the integration tests from the Bitnami Kafka docker-compose to [`testcontainers-rs`](https://github.com/testcontainers/testcontainers-rs), and substantially extend the suite's binding coverage.

## Testcontainers migration

Tests now spin up the broker through `testcontainers-modules`' apache Kafka image, with a single `KafkaContext` shared across tests in a binary via a `tokio::sync::OnceCell` (see `tests/utils/containers.rs`). `cargo test` works with no out-of-band setup beyond a reachable Docker daemon. The CI test matrix fans out across `KAFKA_VERSION = 3.7 / 3.8 / 3.9 / 4.0` and resolves each row to a specific `apache/kafka:<tag>` image.

The migration also reorganised the integration tests into per-area files under `tests/`, with shared helpers in `tests/utils/`. A few assertions had to flex for the new environment:

- `tests/metadata.rs` uses `BROKER_ID = 1` (the image hardcodes it) and drops the `port == 9092` check (testcontainers maps to a random host port).
- `tests/base_producer.rs::test_base_producer_timeout` points the producer at `127.0.0.1:1` so the 100ms `message.timeout.ms` is exercised reliably; Kafka 4.0 with `auto.create.topics.enable=true` was fast enough on CI to deliver the message before the timeout fired.
- `tests/consumer_groups.rs::test_delete_unknown_group` accepts either `GroupIdNotFound` or `NotCoordinator`, since coordinator state on a cold broker is non-deterministic when this test runs first in the binary.

The single-broker testcontainers image also needs `KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR=1` and `KAFKA_TRANSACTION_STATE_LOG_MIN_ISR=1`; without those, the transactions tests hang trying to bring `__transaction_state` up to its default RF of 3.

`testcontainers-modules` is pinned at `0.12.1`. Bumping to `0.13` pulls in `testcontainers 0.25` -> `tonic 0.14` -> rustc 1.88, which is past CI's `rust_version: 1.85` floor. Happy to bump it in a follow-up if you'd rather.

## New integration-test coverage

The testcontainers swap above is a straight port of the pre-existing integration tests. On top of that, this PR adds new tests that extend coverage to binding surfaces the upstream suite didn't reach.

- Producer (`tests/base_producer.rs`, `tests/future_producer.rs`): compression round-trip per codec (`gzip` / `snappy` / `lz4` / `zstd`; zstd is feature-gated and CI runs `cargo test --features zstd` so the codec actually links), idempotent producer with mid-stream `flush()`, `MessageSizeTooLarge` on an oversized `BaseProducer::send`, default partitioner key-to-partition determinism, `BorrowedMessage::headers` round-trip on the consumer side, `Producer::flush` drains in-flight.
- Consumer (`tests/base_consumer.rs`, `tests/stream_consumers.rs`): pre/post rebalance callbacks fire with the expected partition sets, `offsets_for_times` returns the first offset at-or-after, regex subscription assigns only matching topics, cooperative-sticky `rebalance_protocol()` reports `Cooperative` after a group join.
- Admin (`tests/admin.rs`, `tests/consumer_groups.rs`): topic-scoped `alter_configs` readback with `ConfigSource::DynamicTopic`, `create_topics` `Fixed(3)` returns `InvalidReplicationFactor` on a single broker, `delete_groups` returns `NonEmptyGroup` while the consumer is alive and succeeds after the consumer drops.
- Transactions (`tests/transactions.rs`): `send_offsets_to_transaction` end-to-end on a consume-transform-produce loop, `isolation.level` filters interleaved aborted/committed payloads, producer fencing by epoch (second `init_transactions` with the same `transactional.id` fences the first's `commit_transaction`).
- Observability (`tests/base_producer.rs`): `ClientContext::stats` fires on the `statistics.interval.ms` cadence and deserialises into the `Statistics` struct.

## For reviewers

There's a walkthrough of the integration suite at [`tests/README.md`](https://github.com/j7nw4r/rust-rdkafka/blob/j7nw4r/testing-update/tests/README.md) on this branch. It covers the shared-broker pattern, the helper cheatsheet under `tests/utils/`, how to add a new test, the CI matrix, and a "known quirks" section for the non-obvious assertion shapes (the `test_base_producer_timeout` unreachable-broker trick, the `test_delete_unknown_group` either-error tolerance, the 3.7 wakeup-callback baseline). It's the fastest way to context-switch into the new test suite.


j7nw4r:
@fede1024 Sorry had serious health issues so this went by the wayside. I was able to finish this. If it's still wanted please take a look. I added a tests/README.md that may help in understanding.

fede1024:
@j7nw4r I'm really sorry to hear about your health issues. No need to apologize, of course! Health comes first. Thank you for your continued effort on this PR. I'll review asap.

fede1024:
Thank you for the massive testing overhaul! I'm glad that we can entirely remove the Docker configuration and rely on a more standardize approach.

This looks great, I don't have anything in particular to flag. I see that there is at least one test that is still commented out (`test_produce_consume_consumer_nonempty_callback`). It was already commented out before this PR, therefore it's not something we need to fix here (this PR is massive already). However it would be nice to add it back, if we believe it provides any value, at some point in the future in a separate PR.

Also it would be nice at some point to reintroduce memory validation, via ASan. Valgrind has been disabled for a long time, and given that the entire purpose of this library is to interact with a C library, we always need to keep an eye out for potential memory access bugs. Again, not to be addressed in this PR of course.

PR #803, opened 2025-11-05
feat: Update bindings to 2.12.1

Marwes:
Mainly want to see an update to to the re-connection issue on 2.10. Changes the test kafka to bitnamilegacy/kafka as all images were moved
here. Seems to be a risk that these images will also be removed at some point so finding another source for running kafka for testing would be good.

Closes https://github.com/fede1024/rust-rdkafka/pull/789

### DIFF
--- a/rdkafka-sys/build.rs
+++ b/rdkafka-sys/build.rs
@@ -100,6 +100,10 @@ fn main() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Marwes:
Any chance of a release with this?

Issue #800, opened 2025-10-17
Support for librdkafka 2.12.0


sukhmel:
Please consider release with support for librdkafka 2.12.0

It has features for working with new Kafka release 4.0.0, and several bugfixes, including ones for librdkafka 2.11.0 which is currently the latest supported one.

coding-yogi:
➕  It also has [some fixes](https://github.com/confluentinc/librdkafka/commit/5cdc6ec979009a406a19acbcb72b14f74a820f5b) related to OAuth implementation where the static linking of OpenSSL currently fails with `SSL certificate problem: unable to get local issuer certificate (-1)`

DoumanAsh:
@fede1024 Do you plan to make any release soon?

rdettai-sk:
FYI 0.39 is out and it uses librdkafka v0.12.1

PR #805, opened 2025-11-11
Build librdkafka WITH_CURL if curl-static feature is selected

tillrohrmann:
Before librdkafka was only built WITH_CURL if the feature curl was enabled. However,
rdkafka also supports curl-static which statically links libcurl. In this case, librdkafka
should also enable the WITH_CURL feature which is a prerequisite for enabling OIDC oauthbearer.

Additionally, this PR bumps librdkafka to 2.12.1 which allows us to configure the ca certificate location for the statically linked curl library via `https.ca.location` (see also https://github.com/fede1024/rust-rdkafka/issues/800#issuecomment-3426103193).


tillrohrmann:
> You probably also need the changes in `bindings.rs`, see e.g. #803
> 
> I also noticed this curl problem there and I think it might be causing the CI failures.

Awesome. Are you gonna incorporate the curl-static changes in #803 @thalesfragoso? If yes, then #803 should supersede this PR. Feel free to close this PR if this is the case and I'll wait until #803 is merged.

fede1024:
#803 has been merged! I believe this one can be closed now. Feel free to open it again if that's not the case.

Issue #809, opened 2025-11-22
Cannot use Kafka Bootstrap Servers with `localhost:port` syntax


bborutenko:
Summary
- When bootstrap.servers is set to localhost:<port>, the client fails to connect to the broker with BrokerTransportFailure / AllBrokersDown.
- Root cause: localhost resolves to ::1 (IPv6) on the host, while the broker is reachable only via IPv4. librdkafka attempts IPv6 and gets “Network is unreachable.”

Environment
- OS: Linux archlinux 6.17.7
- Kafka: confluentinc/cp-kafka:7.4.4 (Docker)
- Port mapping: host 29092 -> container
- App: apps/analytics (Rust, rdkafka)

Steps to Reproduce
Prerequisites
- Kafka + Zookeeper running in Docker, Kafka exposed on host port 29092 (typical Confluent cp-kafka setup).
- Your app runs on the host (not in Docker).

Steps
1) Create a .env with localhost (this is the trigger)
```
KAFKA_BOOTSTRAP_SERVERS=localhost:29092
KAFKA_GROUP_ID=analytics-consumer
KAFKA_INPUT_TOPIC=events
```

2) Use a consumer config that does NOT force IPv4
```
use rdkafka::config::ClientConfig;
use rdkafka::consumer::{Consumer, StreamConsumer};

pub struct App {
    consumer: StreamConsumer,
}

impl App {
    pub fn new(brokers: &str, group_id: &str, input_topic: &str) -> App {
        let consumer: StreamConsumer = ClientConfig::new()
            .set("group.id", group_id)
            .set("bootstrap.servers", brokers)
            // NOTE: intentionally NOT setting "broker.address.family" to reproduce the issue
            .set("enable.partition.eof", "false")
            .set("session.timeout.ms", "6000")
            .set("enable.auto.commit", "false")
            .create()
            .expect("Consumer creation failed");

        consumer
            .subscribe(&[input_topic])
            .expect("Can't subscribe to specified topic");

        App { consumer }
    }

    pub async fn run(&self) {
        use futures::TryStreamExt;
        let stream_processor = self
            .consumer
            .stream()
            .try_for_each(|borrowed_message| async move {
                let owned = borrowed_message.detach();
                tokio::spawn(async move {
                    let result = async {
                        info!("start processing message");
                        process_message(owned).await
                    }
                    .instrument(span)
                    .await;
                    match result {
                        Ok(_) => ...,
                        Err(error) => ...,
                    };
                });
[...]

PR #810, opened 2025-11-22
Resilient stream example with connection retry

bborutenko:
This change resolves issue #809 
This revised example demonstrates a robust streaming consumer that does not panic on the first transient network error (e.g., BrokerTransportFailure on IPv6 ::1 for localhost). Instead, it restarts the stream with an exponential backoff, allowing librdkafka to retry and fall back to alternative broker addresses (such as 127.0.0.1) when available. The processing of messages remains asynchronous via `tokio::spawn`, while the outer loop handles stream errors gracefully and continuously attempts to reconnect.

### DIFF
--- a/examples/asynchronous_processing.rs
+++ b/examples/asynchronous_processing.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 use clap::{Arg, Command};
 use futures::stream::FuturesUnordered;
 use futures::{StreamExt, TryStreamExt};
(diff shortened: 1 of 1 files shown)


dennisrei:
Not one of the maintainers. But I like the extra information on resolving to ipv4 or ipv6. However, in my opinion I don't think the asynchronous_processing example should be updated with counting connection attempts.

PR #818, opened 2026-01-05
Fix build script for musl builds that require SASL 

tillrohrmann:
This PR fixes the library path for sasl2 and libz since both crates produce the artifacts in /lib. Additionally, it adds the libsasl2.pc to the pkg-config path because librdkafka only uses pkg-config to locate the sasl2 library. With these changes it is possible to build rdkafka for a musl target.

This PR is a rebased version of #806.


tillrohrmann:
@fede1024 could you please help me moving this PR forward. This is currently preventing us from getting rid of our rust-rdkafka fork.

BIHB:
We have the same problem.
Thanks for your PR!
@fede1024 Can you please take a look at it?

fede1024 (approved):

fede1024:
TY!

Issue #665, opened 2024-03-30
`StreamConsumer` not waking


trtt:
This `StreamConsumer` racy behavior can be observed in the recent release, probably after the move to Event-based API. 
Some of the events (stats, logs etc.) are processed internally and are consumed by the `poll` returning `None`. Since `librdkafka` only wakes the queue when it transitions from empty -> non-empty, waker might be not called at any point in the future in this case. 

Event order example:
1) new message event
2) new stats event
3) `poll` processes stats and returns `None`
4) `StreamConsumer` sets up waker and returns `Poll::Pending`
5) waker is never called by librdkafka since the queue never went empty

Probably connected issue: https://github.com/fede1024/rust-rdkafka/issues/638

trtt:
I sketched out a potential solution to the problem: https://github.com/fede1024/rust-rdkafka/pull/666

sukhmel:
I was experiencing something that seems an instance of this issue, if I have `statistics.interval.ms` set and sleep for longer time before consuming messages with `StreamConsumer<LatestStatisicsContext>`.

This problem is gone with `rust-rdkafka@0.37`, thank you for the effort of fixing it. Maybe this issue can be closed?

1990heidou:
Using version 0.35 will result in the same [#638 ](https://github.com/fede1024/rust-rdkafka/issues/638) issue. 
Upgrading to version 0.37 will still encounter the problem of being unable to continue consuming, but not outputting :
**WARN librdkafka: librdkafka: MAXPOLL [thrd:main]: Application maximum poll interval (300000ms) exceeded by ??? ms**
![image](https://github.com/user-attachments/assets/b4f26863-bbab-40a6-872d-212bad56396c)

```rust
fn consumer_config(config_overrides: Option<HashMap<&str, &str>>) -> ClientConfig {
    let mut config = ClientConfig::new();

    config.set("group.id", "group_id");
    config.set("bootstrap.servers", "localhost")
        .set("enable.partition.eof", "false")
        .set("session.timeout.ms", "60000")
        .set("enable.auto.commit", "true")
        .set("max.poll.interval.ms", "60000")
        .set("fetch.wait.max.ms", "50")
        .set("log.thread.name", "false")
        .set("auto.offset.reset", "earliest");

    if let Some(overrides) = config_overrides {
        for (key, value) in overrides {
            config.set(key, value);
        }
    }

    config
}

[...]

PR #840, opened 2026-05-31
Cover split partition queues with the periodic fallback wakeup

davidak09:
## Summary

`StreamConsumer::split_partition_queue` can permanently stall a single partition's stream after a lost wakeup, while every other partition keeps consuming. This adds the periodic fallback wakeup to split partition queues that the main consumer queue already has.

## Background

A parked `MessageStream` is rewoken by an **edge-triggered** "queue non-empty" callback (`native_message_queue_nonempty_cb`) that only fires on an empty → non-empty transition. That edge can be missed (see #665), leaving the stream parked with no pending wakeup.

The main consumer queue is protected against this: `from_config_and_context` spawns a background loop that calls `wakers.wake_all()` every `max.poll.interval.ms / 2`, so a missed edge self-heals within at most that interval.

`split_partition_queue` creates its **own** `WakerSlab` and enables its own non-empty callback — but no equivalent fallback loop is spawned for it. So for split partition queues a missed wakeup is **not** recovered: the stream stays parked indefinitely.

## Impact

Consumers that use `split_partition_queue` (one stream per partition for parallel/independent draining) can have a single partition silently stop consuming for hours/days while sibling partitions on the same consumer keep flowing. Only recreating the consumer clears it. This matches behavior reported downstream in vectordotdev/vector#22006 (one partition's lag climbs unbounded; the consumer is otherwise healthy, no errors logged).

## Fix

When `split_partition_queue` creates the queue's `WakerSlab`, spawn the same fallback loop used for the main queue, waking that queue's wakers every `max.poll.interval.ms / 2`. The loop holds a `Weak<WakerSlab>` and exits once the queue is dropped and its `WakerSlab` is freed, so no shutdown plumbing or changes to `StreamPartitionQueue` are needed. `poll_interval` is now stored on `StreamConsumer` to reuse the value parsed at construction.

Net production change: one file, +10/−16.

## Test

`tests/test_partition_queue_wakeup.rs` — a deterministic, broker-free regression test. It assigns and splits a partition, parks the stream by polling it with a custom waker (empty queue, no broker → no data), and asserts the stream is woken within the timeout. The only thing that can wake it is the fallback loop.

Verified fail → pass: with the fallback loop removed the test times out (`never woken`); with the fix it passes in ~0.5s.

## Validation

- `cargo test --lib` — passes (no broker).
- New regression test — passes with the fix, fails without it.
- Existing consumer integration tests (incl. the other `split_partition_queue` tests) — pass against a real broker (`docker compose up` + `cargo test`), no regression.

## Notes

Relates to #665; addresses the stuck-partition symptom in vectordotdev/vector#22006.

Developed with AI assistance and reviewed by me.


tanganellilore:
@fede1024 
Any news on this PR?

I have simialr issue on vector that use this library.

Thanks

tanganellilore:
I tried to build may custom vector  with this patch branch, but not solve our isse taht probably have different origin.

Thanks

PR #841, opened 2026-06-08
rdkafka-sys: bump librdkafka to 2.14.2 with SSL certificate hot reload

ngirish-code:
## What

Repoints the `rdkafka-sys/librdkafka` submodule to a librdkafka build that adds
**SSL certificate hot reload**, bumping the vendored librdkafka from **2.12.1 → 2.14.2**.

This lets a client pick up rotated TLS certificates without being recreated:
librdkafka rebuilds its `SSL_CTX` from the configured `ssl.*.location` files and
atomically swaps it in for new broker connections.

## Why

Long-lived producers/consumers using short-lived (rotated) TLS certs currently
have to be torn down and rebuilt to load renewed certs. The hot-reload path
avoids that disruption.

## Changes

- `.gitmodules` — point the submodule at the fork carrying the change and track
  its branch.
- `rdkafka-sys/librdkafka` — submodule pinned to `dcd0847a` (librdkafka 2.14.2 +
  cert hot reload).
- `rdkafka-sys/Cargo.toml` — version metadata `4.10.0+2.12.1` → `4.10.0+2.14.2`
  (kept in sync with the runtime `rd_kafka_version_str()`, which
  `version_check.rs` asserts).
- `Cargo.lock` — synced.

No Rust binding changes: the reload is driven entirely by a new config property,
so it works through the existing `ClientConfig::set()` path.

## Usage

```rust
let mut config = ClientConfig::new();
// poll the cert files every 30s and reload on change
config.set("ssl.certificate.refresh.interval.ms", "30000"); 
```
Only file-based certificate sources are supported (in-memory PEM / engine certs are not reloaded).

## Testing

• cargo test -p rdkafka-sys --test version_check passes (reports
  librdkafka version: 2.14.2).
• librdkafka.a builds from the vendored source via the default (mklove) build.

## Notes / caveats

• The submodule currently tracks a personal fork. Before this is suitable for
  a release, the librdkafka change should land upstream and the submodule be
  re-pinned to an upstream commit (restoring the canonical URL).
• The bump pulls in the full 2.12.1 → 2.14.2 upstream delta (30 commits), not
  just the cert-reload commit. The public header surface is unchanged apart from
  the version macro and one new (unbound) rd_kafka_ssl_ctx_reload function, so
  existing bindings remain valid.
• librdkafka also exposes an on-demand rd_kafka_ssl_ctx_reload() API; it is
  intentionally not surfaced in the Rust API in this PR (polling only).


jplatte:
Is there an upstream PR for the hot reloading? I don't think using your fork is reasonable for these bindings.

PR #850, opened 2026-07-20
support aws-lc backend

glebpom:
This PR adds support for using the `aws-lc` backend instead of `openssl`.

`aws-lc` is widely adopted across the Rust ecosystem and is generally preferred over `openssl`. It also provides a FIPS-compliant variant. This PR does not add FIPS support yet, as doing so requires a small change to the main `librdkafka` C codebase - https://github.com/confluentinc/librdkafka/pull/5552

The implementation relies on the `aws-lc-sys` crate for the build process, so it does not introduce any additional system dependencies.

## Dependencies: 

- [x] Support in `aws-lc` for exporting the required variables: https://github.com/aws/aws-lc-rs/pull/1187


glebpom:
@fede1024 could you please take a look? `aws-lc-sys` with the necessary support has been released

PR #854, opened 2026-08-26
Disable zstd-sys bindgen feature

stepancheg:
Keep the remaining zstd-sys defaults (legacy, zdict_builder) so the vendored libzstd build is unchanged.

bindgen makes hermetic compilation (e.g. Bazel) and cross-compilation hard, because it needs libclang at build time. That is a problem even when only a subset of the workspace depends on rdkafka: Cargo feature unification enables the zstd-sys bindgen feature for the whole graph.

rdkafka-sys only uses zstd-sys to build and link libzstd, not its generated Rust bindings, so bindgen is unnecessary here.

Related PR suggesting disabling bindgen from default features: https://github.com/gyscos/zstd-rs/issues/370

AI assistance was used.


stepancheg:
bindgen is no longer default in zstd https://github.com/gyscos/zstd-rs/pull/371 so I think we can close this PR.

